PR #22, opened 2020-02-06
feat(analyzer): add placeholder type

makenowjust:
## Problem

Sometimes one needs to use `@param` without type definition and wants to infer `@param` type from query.

For example one hopes the analyzer infers `@param` type as `STRING` from the below SQL:

```sql
SELECT * FROM Singers WHERE FirstName LIKE @param
```

## Solution

This PR introduces `Placeholder` type and `PlaceholderType` under to `analyzer` package.

- `Placeholder` is *value* type which is typed as `PlaceholderType`.
- `PlaceholderType` is *type* type which can be inferred its type on analyzing phase.

And this implements `PlaceholderType.CoerceTo` sets placeholder type as the target type.

For detailed usage, please see updated `example/analyzer/main.go`.

## TODO

- [ ] Add document
- [ ] Add test
- [ ] Support indirect references like `@param1 = @param2` (is this really needed? Maybe it should be prevented.)

- - -

/cc @kazegusuri

### DIFF
--- a/example/analyze/main.go
+++ b/example/analyze/main.go
@@ -12,7 +12,7 @@ import (
 func main() {
 	// Create a new Parser instance.
 	file := &token.File{
-		Buffer: "SELECT * FROM singers",
+		Buffer: "SELECT * FROM singers WHERE CAST(@param As STRUCT<foo STRING>).foo = FirstName",
 	}
 	p := &parser.Parser{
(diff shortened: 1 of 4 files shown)


git-hulk:
Hello, is this repo still under maintenance?

makenowjust:
Close due to no time to update.

PR #23, opened 2020-03-08
Add concat operator

apstndb:
Add concatenation operator(`||`) which is supported as mul/div precedence in Cloud Spanner.
https://cloud.google.com/spanner/docs/operators?hl=en#concatenation_operator

See also.
https://github.com/google/zetasql/commit/aac17f8352ab620afbf6124d7193ccedc79ec786#diff-c5cedd04cbc91402d05ff54d91e14994R822


MakeNowJust:
Thank you!

PR #27, opened 2020-10-19
Export ParseType

apstndb:
Add exported `(p *Parser) ParseType()`.

ZetaSQL also have a top-level [ParseType function](https://github.com/google/zetasql/blob/2020.09.01/zetasql/parser/parser.cc#L210) and I have a use case of parsing type names(https://github.com/apstndb/execspansql).


MakeNowJust:
There is a use-case, so we can go. Thanks!

apstndb:
Thanks!

PR #26, opened 2020-10-18
Support NUMERIC type

apstndb:
Add preliminary NUMERIC type support.

PR #28, opened 2020-10-19
Fix analyze of ast.NumericLiteral

apstndb:
#26 doesn't have appropriate NUMERIC support in analyzer so fix it.


MakeNowJust:
@apstndb Thanks!

PR #29, opened 2021-02-05
Support parsing generated columns in CREATE TABLE

tyamagu2:
Support parsing [generated columns](https://cloud.google.com/spanner/docs/generated-column/how-to) in CREATE TABLE.

Note that the modified parser parses generated column expression as an ordinary expression, and thus it accepts expressions that are invalid as generated column expression as long as `parseExpr` is able to parse them.


pkg/ast/ast.go
```diff
@@ -1272,10 +1272,24 @@ type ColumnDef struct {
...
 	Null token.Pos // position of "NULL"
 
-	Name    *Ident
-	Type    SchemaType
-	NotNull bool
-	Options *ColumnDefOptions // optional
+	Name          *Ident
+	Type          SchemaType
+	NotNull       bool
+	GeneratedExpr *GeneratedColumnExpr // optional
```
MakeNowJust, line 1278: Please update the above `ColumnDef` doc comment by adding `{{ GeneratedExpr | sqlOpt }}`.

pkg/ast/ast.go
```diff
@@ -1272,10 +1272,24 @@ type ColumnDef struct {
...
+	Options       *ColumnDefOptions    // optional
+}
+
+// GeneratedColumnExpr is generated column expression.
+//
+//     AS ({{.Expr | sql}}) STORED
+type GeneratedColumnExpr struct {
```
MakeNowJust, line 1285: Are `GeneratedColumnExpr.Pos` and `GeneratedColumnExpr.End` missing? They are necessary to implement `Node` interface.

MakeNowJust (changes requested):
Great :tada:

Thanks for noting about parsing. I think such validation is done in analyzing phase instead of parsing, so that's no problem.

pkg/ast/ast.go
tyamagu2: Added `Pos`, `End` and `SQL` to `GeneratedColumnExpr`. Also fixed `ColumnDef.End` to take `GeneratedExpr` into account.

MakeNowJust (approved):

MakeNowJust:
Thank you @tyamagu2!

PR #31, opened 2021-06-21
modify ParseDDL to return nil if query has only comment.(has no idents)

nakatamixi:
Hi!

I use https://github.com/cloudspannerecosystem/yo , and yo use memefish.

I want parse DDL such as 
```
go run ./tools/parse -mode=ddl "# hoge"
2021/06/21 22:37:04 syntax error::1:7: expected token: CREATE, <ident>, but: <eof>

  1:  # hoge
            ^
exit status 1
```

And I changed `parseDDL` to return nil if first token kind is `token.TokenEOF` .
Please review this change.

Thanks!

### DIFF
--- a/pkg/parser/parser.go
+++ b/pkg/parser/parser.go
@@ -1886,6 +1886,8 @@ func (p *Parser) lookaheadType() bool {
 func (p *Parser) parseDDL() ast.DDL {
 	pos := p.Token.Pos
 	switch {
+	case p.Token.Kind == token.TokenEOF:
+		return nil
 	case p.Token.Kind == "CREATE":
 		p.nextToken()
 		switch {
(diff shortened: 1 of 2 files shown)


MakeNowJust:
After getting this PR merged, a `ParseDDLs` result will contain `nil` only when the source has a trailing comment after the last semicolon. Despite this behavior, an alone semicolon will be rejected by the parser. It seems natural to support an empty statement generally.

Another question I have is "why does it need to treat `ParseDDL` as special?" It makes more sense to me that `ParseQuery` and `ParseDML` do so.

nakatamixi:
Thanks for your review.

I'll try to use `ParseDDLs` for my usecase.

nakatamixi:
I can parse my usecase by using `ParseDDLs`.
So I close this.

> Another question I have is "why does it need to treat ParseDDL as special?"

Sorry, there is no reason... I just want to fix my case.
`yo` using `strings.Split(schema, ";")` for parse multiple ddls in strings.
So I'll change yo code.

Thanks.

PR #32, opened 2021-07-14
Support parsing alter table  to add an foreign key

tyamagu2:
* Support parsing alter table to add a foreign key
* Modify `ast.ForeignKey` to implement `ast.Node` interface.

### DIFF
--- a/pkg/parser/parser.go
+++ b/pkg/parser/parser.go
@@ -2030,10 +2030,11 @@ func (p *Parser) parseForeignKey() *ast.ForeignKey {
 		p.nextToken()
 		refColumns = append(refColumns, p.parseIdent())
 	}
-	p.expect(")")
+	rparen := p.expect(")").End
 
 	return &ast.ForeignKey{
 		Foreign:          pos,
+		Rparen:           rparen,
 		Name:             nil,
 		Columns:          columns,
 		ReferenceTable:   refTable,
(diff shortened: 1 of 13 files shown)


pkg/parser/parser.go
```diff
@@ -2278,6 +2279,36 @@ func (p *Parser) parseAlterTable(pos token.Pos) *ast.AlterTable {
 	}
 }
 
+func (p *Parser) parseAlterTableAdd() ast.TableAlternation {
+	lexer := p.Lexer.Clone()
```
MakeNowJust, line 2283: Cloning lexer is a last resort for parsing. When it can be done without `Clone`, we recommend doing so.

@tyamagu2 If you have any idea that it must use `Clone`, please tell me or update a source code comment. If not, please consider implementing this without `Clone`. It seems not so hard.

MakeNowJust (changes requested):
Good catch to support a missing syntax. Thank you!

I wonder if `ALTER TABLE ... DROP FOREIGN KEY` or additional foreign key related syntax are supported in today's Spanner SQL. If they are supported, here is a good place to implement them.

pkg/parser/parser.go
tyamagu2: Maybe we can extract the following part of `parseAddColumn` as `parseColumn` and call it here?

https://github.com/MakeNowJust/memefish/pull/32/files#diff-2a6035c752559ce0fb74ced451e9e4363c01598c669d8f77a46be9450e6f6d87R2314-R2321

MakeNowJust (approved):
LGTM :+1:

tyamagu2:
> I wonder if ALTER TABLE ... DROP FOREIGN KEY or additional foreign key related syntax are supported in today's Spanner SQL. If they are supported, here is a good place to implement them.

Spanner does support `ALTER TABLE DROP CONSTRAINT`. I'll send another PR for that.
https://cloud.google.com/spanner/docs/data-definition-language#alter_table

MakeNowJust:
@tyamagu2 Okay. Thank you so much!

PR #37, opened 2022-03-13
fix(parser): adjust `AND` and `OR` precedence

S-YOU:
Following returns true in spanner query.

```
SELECT TRUE OR FALSE AND FALSE
```

But current AST returns following

```go
       Expr: &ast.BinaryExpr{
          Op:   "AND",
          Left: &ast.BinaryExpr{
            Op:   "OR",
            Left: &ast.BoolLiteral{
              ValuePos: 7,
              Value:    true,
            },
            Right: &ast.BoolLiteral{
              ValuePos: 15,
              Value:    false,
            },
          },
          Right: &ast.BoolLiteral{
            ValuePos: 25,
            Value:    false,
          },
        },
      },
```

Which will evaluate to `false`, which is not correct I think and spec says AND has higher precedence.

https://cloud.google.com/spanner/docs/reference/standard-sql/operators

Order of Precedence | Operator | Input Data Types | Name | Operator Arity
-- | -- | -- | -- | --
11 | AND | BOOL | Logical AND | Binary
12 | OR | BOOL | Logical OR | Binary

### DIFF
--- a/pkg/ast/sql.go
+++ b/pkg/ast/sql.go
@@ -18,7 +18,8 @@ const (
 	precBitOr
 	precComparison
 	precNot
-	precAndOr
+	precAnd
+	precOr
 )
 
 func exprPrec(e Expr) prec {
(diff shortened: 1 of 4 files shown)


makenowjust:
@S-YOU Thank you!

PR #39, opened 2022-04-07
Add support for the CHECK constraint

tyamagu2:
closes https://github.com/makenowjust/memefish/issues/38


pkg/ast/ast.go
```diff
@@ -1263,7 +1263,7 @@ type CreateTable struct {
 	Name              *Ident
 	Columns           []*ColumnDef
 	PrimaryKeys       []*IndexKey
-	ForeignKeys       []*ForeignKey
+	TableConstraints  []*TableConstraint       // optional
```
makenowjust, line 1266: I'm not sure this `// optional` comment is necessary. Contrary, there is `// len(Whens) > 0` comment, so we assumed all slice fields without comment contain an empty slice.

pkg/ast/ast.go
```diff
@@ -1404,16 +1434,14 @@ type AddColumn struct {
 	Column *ColumnDef
 }
 
-// AddForeignKey is ADD FOREIGN KEY clause in ALTER TABLE.
-//
-//     ADD {{.ForeignKey | sql}}
-type AddForeignKey struct {
+// AddTableConstraint is ADD table_constraint clause in ALTER TABLE.
```
makenowjust, line 1437:
```suggestion
// AddTableConstraint is ADD table_constraint clause in ALTER TABLE.
//
//     ADD {{.TableConstraint}}
```

makenowjust (changes requested):
Implementation seems good. I was curious about some comments. Please see above.

pkg/ast/ast.go
tyamagu2, line 1266:
```suggestion
	TableConstraints  []*TableConstraint
```

tyamagu2: OK, makes sense.

makenowjust (approved):
LGTM. Thank you.

PR #41, opened 2022-04-11
Fix test result files

tyamagu2:
Some test result files added by https://github.com/makenowjust/memefish/pull/40 had unintentional format changes, making the CI fail.


makenowjust:
Thanks!

PR #42, opened 2022-04-11
Fix CI triggers

makenowjust:
This PR fix CI triggers to run on each pull requests.


makenowjust:
Seems good.

スクリーンショット 2022-04-11 16 44 59

PR #47, opened 2022-06-14
Support IF conditional expression syntax

### LINKED ISSUE #46: Support IF conditional expression syntax
### Search before asking

- [X] I had searched in the [issues](https://github.com/makenowjust/memefish/issues) and found no similar issues.

### Motivation

Spanner supported the `IF` conditional expression, but `memefish` can't parse the syntax in alter table SQL. As a user, I would be happy to see it can support syntax as more as possible. For more detail can see: [conditional_expressions#if](https://cloud.google.com/spanner/docs/reference/standard-sql/conditional_expressions#if)

### Solution

Parse the IF expression which is described in [conditional_expressions#if](https://cloud.google.com/spanner/docs/reference/standard-sql/conditional_expressions#if)

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

git-hulk:
This closes #46


makenowjust:
@git-hulk No. I will fix it. Thank you!

makenowjust:
@git-hulk I'm too busy to have no time to review now. Please wait. Sorry.

git-hulk:
@makenowjust Thanks for your kind reply. No hurry, can have a look after you have the free time, I'm very appreciate your great work.

pkg/ast/ast.go
```diff
@@ -935,6 +936,18 @@ type CaseElse struct {
 	Expr Expr
 }
 
+// IfExpr is IF expression node.
+//
+//     IF ( {{.Expr | sql}}, {{.TrueResult | sql}}, {{.FalseResult | sql}} )
+type IfExpr struct {
```
makenowjust, line 942: Why is this new struct necessary? Could we reuse `CallExpr`?

pkg/parser/parser.go
```diff
@@ -948,9 +948,29 @@ func (p *Parser) parseTableSampleSize() *ast.TableSampleSize {
 // ================================================================================
 
 func (p *Parser) parseExpr() ast.Expr {
+	if p.Token.Kind == "IF" {
```
makenowjust, line 951: I'm not sure this is true. For example, is `IF(X < Y, 1, 2) + 1` invalid?

pkg/parser/parser.go
```diff
@@ -1373,6 +1393,8 @@ func (p *Parser) parseLit() ast.Expr {
 		return p.parseParam()
 	case "CASE":
 		return p.parseCaseExpr()
+	case "IF":
+		return p.parseIFExpr()
```
makenowjust, line 1397: Ditto here. `false OR IF(X < Y, true, false)` is valid, but `IF(X < Y, true, false) OR false` is invalid. Is it really?

I guess we should call `parseIfExpr` from `parseLit` instead.

pkg/parser/parser.go
```diff
@@ -948,9 +948,29 @@ func (p *Parser) parseTableSampleSize() *ast.TableSampleSize {
 // ================================================================================
 
 func (p *Parser) parseExpr() ast.Expr {
+	if p.Token.Kind == "IF" {
+		return p.parseIFExpr()
+	}
 	return p.parseOr()
 }
 
+func (p *Parser) parseIFExpr() ast.Expr {
```
makenowjust, line 957: Rename `parseIfExpr`.

makenowjust (changes requested):
I have some questions to your fix. Please check out my comments.

pkg/parser/parser.go
git-hulk: Good catch

git-hulk:
@makenowjust I fix those comment issues, can help to have a look again when you have free time.

makenowjust:
The fundamental bug is that `IF` is treated as the keyword instead of an identifier. You can revert all fixes, and remove `"IF"` from [the keyword list](https://github.com/makenowjust/memefish/blob/c7cb34f1544c6f6a85aa48107bdb76725e493738/pkg/token/keywords.go#L56) for this issue. This fix seems canonical to me.

pkg/ast/ast.go
git-hulk: OK, I got your point. I said the `IfExpr` is different from `CallExpr`, coz if it's a expression not a function.

git-hulk:
> The fundamental bug is that `IF` is treated as the keyword instead of an identifier. You can revert all fixes, and remove `"IF"` from [the keyword list](https://github.com/makenowjust/memefish/blob/c7cb34f1544c6f6a85aa48107bdb76725e493738/pkg/token/keywords.go#L49) for this issue. This fix seems canonical to me.

OK, I'll rollback those commits and remove the `IF` keyword. Thanks @makenowjust

git-hulk:
Thanks @makenowjust review

PR #48, opened 2022-07-05
Allow using default expression in alter column

git-hulk:


Joker-hh (approved):

pkg/ast/ast.go
```diff
@@ -1584,9 +1584,10 @@ type AlterColumn struct {
 	Alter token.Pos // position of "ALTER" keyword
 	Null  token.Pos // position of "NULL"
 
-	Name    *Ident
-	Type    SchemaType
-	NotNull bool
+	Name        *Ident
+	Type        SchemaType
+	NotNull     bool
+	DefaultExpr *ColumnDefaultExpr
```
makenowjust, line 1590: Please update the above doc comment, and `End` method.

pkg/ast/ast.go
```diff
@@ -1598,8 +1599,9 @@ type AlterColumnSet struct {
 
 	Alter token.Pos // position of "ALTER" keyword
 
-	Name    *Ident
-	Options *ColumnDefOptions
+	Name        *Ident
+	Options     *ColumnDefOptions
+	DefaultExpr *ColumnDefaultExpr
```
makenowjust, line 1604: Ditto.

makenowjust (changes requested):
Thank you!

To update doc comment and `End` method is missing. Please do that.

pkg/ast/ast.go
git-hulk: Sure, my bad.

pkg/ast/ast.go
```diff
@@ -1576,11 +1576,10 @@ type SetOnDelete struct {
 
 // AlterColumn is ALTER COLUMN clause in ALTER TABLE.
 //
-//     ALTER COLUMN {{.Name | sql}} {{.Type | sql}} {{if .NotNull}}NOT NULL{{end}}
+//     ALTER COLUMN {{.Name | sql}} {{.Type | sql}} {{if .NotNull}}NOT NULL{{end}} {{.DefaultExpr | sql}}
```
makenowjust, line 1579: It should be `{{.DefaultExpr | sqlOpt}}` instead.

pkg/ast/ast.go
```diff
@@ -1592,10 +1591,10 @@ type AlterColumn struct {
 
 // AlterColumnSet is ALTER COLUMN SET clause in ALTER TABLE.
 //
-//     ALTER COLUMN {{.Name | sql}} SET {{.Options | sql}}
+//     ALTER COLUMN {{.Name | sql}} SET {{.Options | sql}} || {{.DefaultExpr | sql}}
```
makenowjust, line 1594: It should be `{{if .Options}}{{.Options | sql}}{{else}}{{.DefaultExpr | sql}}{{end}}`.

makenowjust (changes requested):

pkg/ast/ast.go
git-hulk: Fixed, sorry that I didn't understand the doc syntax exactly. Any guideline for this?

makenowjust (approved):

pkg/ast/ast.go
makenowjust: Syntax is Go's `text/template`, and the three extra filters `sql`, `sqlOpt`, and `sqlJoin` are defined.

- `sql` calls `SQL` method and returns its result.
- `sqlOpt` is similar to `sql`, but it returns empty string when the input is `nil`.
- `sqlJoin` calls `SQL` method against each element of the input slice, then returns the string joining results with the argument as its separator.

git-hulk:
@makenowjust Should I add more test case for `End` function to make codecov happy?

makenowjust:
@git-hulk It sounds good.

PR #49, opened 2022-07-22
Allow trailing semicolon in SQL

git-hulk:


makenowjust:
@git-hulk When I designed the parser, I considered semicolons should be proceeded with before parsing. Do you have any reason that they should be proceeded on parsing?

git-hulk:
@makenowjust Thanks for your quick reply. We use `memefish` to format the Spanner SQL
and users may copy their SQL statements which ends with semicolons. Does it have any
side effect if we allow the trailing semicolon?

makenowjust:
@git-hulk My position: Semicolon is just a separator of a statement, thus this is not a part of a statement. Your change may make it difficult that use semicolons in such a correct way.

You should use `ParseStatements` instead in this case.

git-hulk:
@makenowjust Thanks for your kindly reply, let’s close this PR.

PR #52, opened 2023-03-23
Parse token comments in lexer

kazegusuri:
* Add array of TokenComment to Token
* Parse comments as TokenComment in lexer
* Token.Space has spaces only in this change
   * The leading spaces for comments are saved as TokenComment.Space

I would like to parse statements without building an AST by parser. I found the current lexer can parse statements with skipping comments. However after getting tokens as the result, it's difficult to build a statement by the tokens without comments because Token.Space contains the comments.

If lexer can parse comments correctly, we will be able to use the comments from AST later.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,6 +4,7 @@ go 1.12
 
 require (
 	github.com/MakeNowJust/heredoc/v2 v2.0.1
+	github.com/google/go-cmp v0.5.9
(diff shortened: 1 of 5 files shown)


makenowjust (approved):
Looks good. Thank you.

kazegusuri:
Can I merge this PR? I just confirm if it's need to be reviews after a small change.

makenowjust:
Sorry for late. I merged it. Thank you.

PR #55, opened 2023-04-03
update go to 1.19

kazegusuri:
Update go version in go.mod to v1.19 because go 1.19 is already used in CI.


makenowjust:
Seems good. Thank you.

PR #58, opened 2023-08-19
support CHANGE STREAM

nktks:
This PR adds support for [CHANGE STREAM statements](https://cloud.google.com/spanner/docs/reference/standard-sql/data-definition-language#change_stream_statements)


ast/ast.go
```diff
@@ -1489,6 +1495,39 @@ type AlterTable struct {
...
+	//          .Options.end
+	//       {{ else }}
+	//         {{ if len(.Watch) > 0 }}
+	//           .Watch[len(.Watch)-1].end
+	//         {{ else if .WatchAll }}
+	//           "ALL".end
+	//         {{ else }}
+	//           .Name.end
+	//         {{ end }}
+	//       {{ end }}
```
makenowjust, line 1522: We have not used Go's template-like syntax for `pos` and `end`. If you want to use some rich features here, I think they are code smells.

ast/ast.go
```diff
@@ -1662,6 +1701,49 @@ type CreateRole struct {
...
+//
+//	{{.TableName}} {{if .Columns}}({{ .Columns | sqlJoin ","}}){{end}}
+type ChangeStreamWatch struct {
+	// pos = TableNamePos
+	// end = {{if .Columns }}{{ Rparen + 1 }}{{else}}.TableName.end{{end}}
+
+	TableNamePos   token.Pos // position of this watch definition
+	TableName      *Ident
+	Columns        []*Ident
+	Lparen, Rparen token.Pos // position of "(" and ")"
```
makenowjust, line 1743: `Lparen` is unnecessary because it is used for neither `pos` nor `end`.

ast/ast.go
```diff
@@ -1662,6 +1701,49 @@ type CreateRole struct {
...
+}
+
+// Watch is tables and columns that are watched by change streams.
+//
+//	{{.TableName}} {{if .Columns}}({{ .Columns | sqlJoin ","}}){{end}}
+type ChangeStreamWatch struct {
+	// pos = TableNamePos
+	// end = {{if .Columns }}{{ Rparen + 1 }}{{else}}.TableName.end{{end}}
+
+	TableNamePos   token.Pos // position of this watch definition
```
makenowjust, line 1740: `TableName` has its position, so this field is unnecessary.

ast/ast.go
makenowjust: Also, `token.Pos` fields are put into the top of fields and separated by a newline to other fields. See other definitions.

makenowjust (changes requested):
Thank you for your contribution. It is welcome to support new syntaxes.

Duplication of `Watch` and `WatchAll` seems bad to me. Because they are mutual with each other, it is good to introduce a new interface for a `WATCH` clause.

Some parts are not conventional. Please fix them.

ast/ast.go
nktks: I just would like to write a comment for description of pos or end like the existing code comments,
What is the syntax as in https://github.com/cloudspannerecosystem/memefish/blob/86cbd88693043a3c1ed99d69dbcdd814b187a985/ast/ast.go#L385-L386 ?

nktks:
@makenowjust 
Thanks for your review quickly.

> Duplication of Watch and WatchAll

I investigated that the below two statements are different.(Although this behavior isn't docummented cleary.)
`CREATE CHANGE STREAM change_stream_name`: Just create change stream resource, not watching everything.
`CREATE CHANGE STREAM change_stream_name For ALL`: create change stream resource watching all.

So I added WatchAll to distinguish them.

ast/pos.go
```diff
@@ -659,6 +659,23 @@ func (c *CreateRole) End() token.Pos {
...
+
+func (c *CreateChangeStream) End() token.Pos {
+	if len(c.Options) > 0 {
+		return c.Options[len(c.Options)-1].End() + 1
+	}
+	if len(c.Watch) > 0 {
+		return c.Watch[len(c.Watch)-1].Rparen + 1
+	}
+	if c.WatchAll {
+		return c.Name.End() + token.Pos(len("FOR ALL"))
```
git-hulk, line 674: It seems incorrect to count the position by plus the `len("FOR ALL")` since there may be more than one space between FOR and ALL. Same as other places.

nktks:
I commited
 https://github.com/cloudspannerecosystem/memefish/pull/58/commits/116521d369d6ad720e9057bb41d9edef05ef0dec
so that:
- devide watch, options  ast
- fix `End()` implementation to save pos needed from `End()` functions.

nktks:
I can delete `WatchAll` ( We can set`ALL` case when WatchTable length is 0) for now, I'll fix it.

git-hulk (approved):

makenowjust (changes requested):
There are a lot of issues in this Pull Request, and it is hard to review all of them at once.

Comments:

- Please match coding styles and comments to the existing code base.
- Don't develop your own notation or expression. Please check the existing code base before it.
  * If the notation you want is not found, it's possible that what you want to do is wrong in many cases.
  * Even though you want a new notation, you must explain the reason for its necessity in the Pull Request description.
- AST definitions should be based on the structure described in Spanner's documentation.
  * In this view, we cannot accept the AST definitions you defined except for `DropChangeStream` so far.

Then, I want you to do the following.

- Please rethink the definition of `CreateChangeStream` and fix it.

(`AlterChangeStream` is more complex than `CreateChangeStream`, so let you rethink later.)

makenowjust:
@nktks Could you have any responses?

nktks:
Thanks for your comment and sorry for delay...

According to your comment, I'll try to fix below when I have time.

- modify `CreateChangeStream`, `AlterChangeStream` struct to match [Google defined statements](https://cloud.google.com/spanner/docs/reference/standard-sql/data-definition-language#change_stream_statements)
- fix comment to match definition written in https://github.com/cloudspannerecosystem/memefish/blob/1b120de904be1345085ee862d3d061476dc2db2f/ast/ast.go#L1-L31

ast/ast.go
```diff
@@ -1766,6 +1802,81 @@ type CreateIndex struct {
...
+	// end = TableName.end || Rparen + 1
+	TableName *Ident
+	Columns   []*Ident
+	Rparen    token.Pos // position of ")"
+}
+
+type ChangeStreamOptions struct {
+	Exprs  []Expr
+	Rparen token.Pos // position of ")"
+}
```
makenowjust, line 1850: You should not define incomplete nodes.

My design is the following:

```golang
// ChangeStreamOptions is OPTIONS clause node in CREATE CHANGE STREAM.
//
//	OPTIONS ({{.Records | sqlJoin ","}})
type ChangeStreamOptions struct {
	// pos = Options
	// end = Rparen

	Options token.Pos // position of "OPTIONS" keyword
	Rparen  token.Pos // position of ")"

	Records []*ChangeStreamOptionsRecord
}

// ChangeStreamOptionsRecord is OPTIONS record node.
//
//	{{.Key | sql}}={{.Expr | sql}}
type ChangeStreamOptionsRecord struct {
	// pos = Key
	// end = Value.end

	Key   *Ident
	Value Expr
}
```

makenowjust:
The overall design seems good to me. Please correct some coding styles.

ast/ast.go
```diff
@@ -313,6 +313,25 @@ type InsertInput interface {
 func (ValuesInput) isInsertInput()   {}
 func (SubQueryInput) isInsertInput() {}
 
+// ChangeStreamWatch represents watch statement in CREATE/ALTER CHANGE STREAM statement.
+type ChangeStreamWatch interface {
```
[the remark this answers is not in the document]
nktks: I'm not sure the word `watch` often suggests a clock...
It seems that [google uses watch in other spanner document](https://cloud.google.com/spanner/docs/change-streams#implicit-watching).

But indeed, in [document for DDL](https://cloud.google.com/spanner/docs/reference/standard-sql/data-definition-language#change_stream_statements), it seems that `watch` doesn't appear.
So I'll change to `ChangeStreamFor`. Thank you!

ast/sql.go
```diff
@@ -1,6 +1,8 @@
 package ast
 
 import (
+	"fmt"
```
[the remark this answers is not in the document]
nktks: I understand we don't use `fmt` in memefish for performance.
So I'll fix not to use it.

ast/ast.go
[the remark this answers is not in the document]
makenowjust: "watching" is not "watch". Google does not use "watch" in the your sense.

parser.go
```diff
@@ -2519,7 +2669,7 @@ func (p *Parser) tryParseInterleaveIn() *ast.InterleaveIn {
 }
 
 func (p *Parser) parseAlterTable(pos token.Pos) *ast.AlterTable {
-	p.expectKeywordLike("TABLE")
+	p.nextToken()
```
makenowjust, line 2672: What is this change?

nktks: Due to this change https://github.com/cloudspannerecosystem/memefish/pull/58/files/9b3a4cb38877bf8fe17cb64fba73d1ae297ab648#diff-83eb8e32639d01cf443d6d8bde24c1c8be78766090d8c5f8586c36250cfedca6R1997-R2002 ,
I changed to check token after `ALTER` token in `parseDDL`.
When next token is `TABLE`, then we call `parseAlterTable()` with `p.Token.Pos` points `TABLE`, when next tokens are `CHANGE STREAM`, then we call `parseAlterChangeStream()`. 

 Is this change acceptable?

makenowjust: No. If `p.Token` is set to a specific value when `p.parseAlterTable` is called, then `p.expectKeywordLike` should all the more be used to make it clear.

ast/ast.go
```diff
@@ -1766,6 +1804,106 @@ type CreateIndex struct {
...
+	// end = Value.end
+
+	Key   *Ident
+	Value Expr
+}
+
+// ChangeStreamAlternationSetFor is SET FOR tables node in ALTER CHANGE STREAM
+//
+//	SET FOR {{.For | sql }}
+type ChangeStreamAlternationSetFor struct {
```
makenowjust, line 1879: The name `ChangeStreamAlternationSetFor` seems redundant. `ChangeStreamSetFor` is better. Other `ChangeStreamAlternationXXX` types are ditto.

makenowjust (changes requested):

parser.go
```diff
@@ -1988,11 +1988,23 @@ func (p *Parser) parseDDL() ast.DDL {
 			return p.parseCreateIndex(pos)
 		case p.Token.IsKeywordLike("ROLE"):
 			return p.parseCreateRole(pos)
+		case p.Token.IsKeywordLike("CHANGE"):
+			p.nextToken()
+			p.expectKeywordLike("STREAM")
+			return p.parseCreateChangeStream(pos)
```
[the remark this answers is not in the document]
nktks: OK. Currently after `ALTER CHANGE` should only be `STREAM`, so I'll fix as above. Thank you!

ast/ast.go
nktks: Does this https://github.com/cloudspannerecosystem/memefish/pull/58/files/2107b868eb59e18a8c7d9d6c8f3694e4bf8c7c11#diff-9dcefbf65d58a909b6cc7332b7985d070382bf920a4b13c7acea4ee91ecdf5acR326 interface `ChangeStreamAlternation` need to change name too?

I'm not sure which name is tasteful...

nktks: At first, I fixed only struct name.
If we also need to fix interface name, please pointing out.

makenowjust: You don't need to fix the interface name. `ChangeStreamAlternation` is good.

ast/ast.go
```diff
@@ -307,6 +313,25 @@ type InsertInput interface {
 func (ValuesInput) isInsertInput()   {}
 func (SubQueryInput) isInsertInput() {}
 
+// ChangeStreamFor represents for statement in CREATE/ALTER CHANGE STREAM statement.
```
makenowjust, line 316:
```suggestion
// ChangeStreamFor represents FOR clause in CREATE/ALTER CHANGE STREAM statement.
```

ast/ast.go
```diff
@@ -1766,6 +1804,109 @@ type CreateIndex struct {
...
+//	SET FOR {{.For | sql }}
+type ChangeStreamSetFor struct {
+	// pos = Set
+	// end = For.end
+
+	Set token.Pos // position of "SET" keyword
+	For ChangeStreamFor
+}
+
+// ChangeStreamDropForAll is DROP FOR ALL node in ALTER CHANGE STREAM
```
makenowjust, line 1890:
```suggestion
// ChangeStreamDropForAll is DROP FOR ALL node in ALTER CHANGE STREAM
//
//	DROP FOR ALL
```

ast/sql.go
```diff
@@ -968,6 +968,77 @@ func (c *CreateIndex) SQL() string {
...
+}
+
+func (a ChangeStreamSetFor) SQL() string {
+	return "SET " + a.For.SQL()
+}
+func (a ChangeStreamDropForAll) SQL() string { return "DROP FOR ALL" }
+func (a ChangeStreamSetOptions) SQL() string {
+	sql := "SET " + a.Options.SQL()
+	return sql
+}
```
makenowjust, line 1020:
```suggestion
func (a ChangeStreamSetFor) SQL() string {
	return "SET " + a.For.SQL()
}

func (a ChangeStreamDropForAll) SQL() string {
	return "DROP FOR ALL"
}

func (a ChangeStreamSetOptions) SQL() string {
	return "SET " + a.Options.SQL()
}
```

ast/sql.go
```diff
@@ -968,6 +968,77 @@ func (c *CreateIndex) SQL() string {
...
+		sql += "("
+		for i, id := range c.Columns {
+			if i > 0 {
+				sql += ", "
+			}
+			sql += id.SQL()
+		}
+
+		sql += ")"
+	}
```
makenowjust, line 1034:
```suggestion
		}
		sql += ")"
	}
```

ast/pos.go
```diff
@@ -679,6 +701,20 @@ func (c *CreateRole) End() token.Pos { return c.Name.End() }
...
+func (a *AlterChangeStream) Pos() token.Pos      { return a.Alter }
+func (a *AlterChangeStream) End() token.Pos      { return a.ChangeStreamAlternation.End() }
+func (a *ChangeStreamSetFor) Pos() token.Pos     { return a.Set }
+func (a *ChangeStreamSetFor) End() token.Pos     { return a.For.End() }
+func (a *ChangeStreamDropForAll) Pos() token.Pos { return a.Drop }
+func (a *ChangeStreamDropForAll) End() token.Pos { return a.All + token.Pos(len("ALL")) }
+func (a *ChangeStreamSetOptions) Pos() token.Pos { return a.Set }
+func (a *ChangeStreamSetOptions) End() token.Pos {
+	return a.Options.Rparen + token.Pos(len(")"))
+}
```
makenowjust, line 716:
```suggestion
func (a *AlterChangeStream) Pos() token.Pos      { return a.Alter }
func (a *AlterChangeStream) End() token.Pos      { return a.ChangeStreamAlternation.End() }

func (a *ChangeStreamSetFor) Pos() token.Pos     { return a.Set }
func (a *ChangeStreamSetFor) End() token.Pos     { return a.For.End() }

func (a *ChangeStreamDropForAll) Pos() token.Pos { return a.Drop }
func (a *ChangeStreamDropForAll) End() token.Pos { return a.All + 3 }

func (a *ChangeStreamSetOptions) Pos() token.Pos { return a.Set }
func (a *ChangeStreamSetOptions) End() token.Pos { return a.Options.Rparen + 1 }
```

ast/pos.go
```diff
@@ -748,6 +784,14 @@ func (s *SizedSchemaType) End() token.Pos { return s.Rparen + 1 }
 func (a *ArraySchemaType) Pos() token.Pos { return a.Array }
 func (a *ArraySchemaType) End() token.Pos { return a.Gt + 1 }
 
+func (c *ChangeStreamForTable) End() token.Pos {
+	if len(c.Columns) == 0 {
+		return c.TableName.End()
+	}
+
+	return c.Rparen + 1
+}
```
makenowjust, line 793:
```suggestion
func (c *ChangeStreamForTable) End() token.Pos {
	if len(c.Columns) == 0 {
		return c.TableName.End()
	}
	return c.Rparen + 1
}
```

parser.go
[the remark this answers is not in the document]
makenowjust: Please do that for other `parse(Alter|Drop)ChangeStream`.

makenowjust:
@nktks https://github.com/cloudspannerecosystem/memefish/pull/58#discussion_r1334166420

nktks:
Oh, I missed it.

makenowjust (approved):
LGTM. Thank you!

nktks:
Thank you!

PR #59, opened 2023-08-22
support  GENERATE_ARRAY

shiv3:
This PR adds support for GENERATE_ARRAY


makenowjust:
Memefish already supports `Call` expression. Why is this specialization necessary?

shiv3:
Thank you. I found that our requirements can be solved with `Call`

PR #50, opened 2022-11-06
Feature/fgac

nktks:
Hello.

This PR enable to support [ fine-grained access control](https://cloud.google.com/spanner/docs/fgac-about) syntax.

Since test failed at my `go1.19.2 darwin/amd64` pc, I updated [golang.org/x/sys](https://github.com/makenowjust/memefish/pull/50/commits/36b8ea46bf6edf4c6cd767f38128e0794fef90ca)
and update golangci-lint version according to https://github.com/golangci/golangci-lint/issues/3269#issuecomment-1265456166 .
Also to use go install, I update CI go version to 1.19.

```
go version
go version go1.19.2 darwin/amd64
[2022-11-10T21:03:04+09:00]/Users/t-nakata/go/src/github.com/makenowjust/memefish (master)(k:gke_mercari-jp-citadel-dev_asia-northeast1_citadel-2g-dev-tokyo-01)(gcp:mercari-jp-citadel-dev) $
> go test ./...
# golang.org/x/sys/unix
../../../../pkg/mod/golang.org/x/sys@v0.0.0-20200202164722-d101bd2416d5/unix/syscall_darwin.1_13.go:25:3: //go:linkname must refer to declared function or variable
../../../../pkg/mod/golang.org/x/sys@v0.0.0-20200202164722-d101bd2416d5/unix/zsyscall_darwin_amd64.1_13.go:27:3: //go:linkname must refer to declared function or variable
../../../../pkg/mod/golang.org/x/sys@v0.0.0-20200202164722-d101bd2416d5/unix/zsyscall_darwin_amd64.1_13.go:40:3: //go:linkname must refer to declared function or variable
../../../../pkg/mod/golang.org/x/sys@v0.0.0-20200202164722-d101bd2416d5/unix/zsyscall_darwin_amd64.go:28:3: //go:linkname must refer to declared function or variable
../../../../pkg/mod/golang.org/x/sys@v0.0.0-20200202164722-d101bd2416d5/unix/zsyscall_darwin_amd64.go:43:3: //go:linkname must refer to declared function or variable
../../../../pkg/mod/golang.org/x/sys@v0.0.0-20200202164722-d101bd2416d5/unix/zsyscall_darwin_amd64.go:59:3: //go:linkname must refer to declared function or variable
../../../../pkg/mod/golang.org/x/sys@v0.0.0-20200202164722-d101bd2416d5/unix/zsyscall_darwin_amd64.go:75:3: //go:linkname must refer to declared function or variable
[...]

PR #66, opened 2023-08-28
Rework fine-grained access control syntax implementations

makenowjust:
This PR mostly reverts #50 and adds correct implementations for fine-grained access control syntax (`GRANT` and `REVOKE` statements).

We introduces two interfaces. `Privilege` represents a privilege on certain object (e.g. TABLE, VIEW, etc). `TablePrivilege` represents a privilege on TABLE.


makenowjust:
`SELECT ON CHANGE STREAM` privilege is not supported so far. It is intentional because #58 is working in progress.

nktks (approved):

ast/sql.go
```diff
@@ -941,64 +950,108 @@ func (d *DropRole) SQL() string {
...
-			sql += ", " + string(priv.Name)
-			if priv.Columns != nil {
-				sql += "(" + CommaSeparatedSQL(priv.Columns) + ")"
-			}
-		}
-		sql += " ON TABLE " + CommaSeparatedSQL(g.TableNames)
-	} else {
-		sql += "ROLE " + CommaSeparatedSQL(g.GrantRoleNames)
+	sql += g.Privilege.SQL()
+	sql += " TO ROLE " + g.Roles[0].SQL()
```
git-hulk, line 954: Not sure if we should use `Strings.Join` to simplify those concat statements.

makenowjust: Spanner SQL is not ours but is Google's. That is, we cannot know the complete picture of it. Without the complete view, we cannot find common parts of the code actually. If we find code that looks like common parts but Google disagrees with it, the commonalities we introduced are wasted. Therefore, we did not use common utilities in `ast` and `parser`.

git-hulk: Sorry for the wrong words. What I mean is that maybe we can use `Strings.Join` to concat the SQL instead of for-loop.

git-hulk: Thank you!

PR #71, opened 2023-08-30
add support for missing IF (NOT) EXISTS clauses

toga4:
This PR adds support for following DDL syntax.

- [CREATE INDEX IF NOT EXISTS](https://cloud.google.com/spanner/docs/reference/standard-sql/data-definition-language#create-index)
- [ALTER TABLE ADD COLUMN IF NOT EXISTS](https://cloud.google.com/spanner/docs/reference/standard-sql/data-definition-language#alter_table)
- [DROP INDEX IF EXISTS](https://cloud.google.com/spanner/docs/reference/standard-sql/data-definition-language#drop-index)

### DIFF
--- a/parser.go
+++ b/parser.go
@@ -2730,10 +2730,17 @@ func (p *Parser) parseDropTable(pos token.Pos) *ast.DropTable {
 
 func (p *Parser) parseDropIndex(pos token.Pos) *ast.DropIndex {
 	p.expectKeywordLike("INDEX")
+	ifExists := false
+	if p.Token.IsKeywordLike("IF") {
+		p.nextToken()
+		p.expect("EXISTS")
+		ifExists = true
+	}
 	name := p.parseIdent()
 	return &ast.DropIndex{
-		Drop: pos,
-		Name: name,
+		Drop:     pos,
+		IfExists: ifExists,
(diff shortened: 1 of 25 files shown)


parser.go
```diff
@@ -2451,6 +2451,15 @@ func (p *Parser) parseCreateIndex(pos token.Pos) *ast.CreateIndex {
...
 
 	p.expectKeywordLike("INDEX")
+
+	ifNotExists := false
+	if p.Token.IsKeywordLike("IF") {
+		p.nextToken()
+		p.expect("NOT")
+		p.expect("EXISTS")
+		ifNotExists = true
+	}
```
toga4, line 2461: This part would appears every time when parsing IF NOT EXISTS option. So it might be better to extract to a function like `func (*Parser) parseIfNotExists() bool`. What do you think?

makenowjust: It sounds good. `IF EXISTS` also appears twice, so we can extract this too. I think a better place to put them would be behind the DDL section.

toga4: Thanks for suggestion. Same here.
Done. 10becb3

makenowjust (approved):
Thank you!

PR #79, opened 2024-05-29
 Add support of 'OR UPDATE|IGNORE' syntax for INSERT DML

git-hulk:
see: https://cloud.google.com/spanner/docs/reference/standard-sql/dml-syntax#insert-ignore

### DIFF
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -1633,7 +1633,7 @@ type AlterTable struct {
 
 	Alter token.Pos // position of "ALTER" keyword
 
-	Name             *Ident
+	Name            *Ident
 	TableAlteration TableAlteration
 }
 
@@ -1646,7 +1646,7 @@ type AlterIndex struct {
 
 	Alter token.Pos // position of "ALTER" keyword
 
-	Name             *Ident
+	Name            *Ident
 	IndexAlteration IndexAlteration
 }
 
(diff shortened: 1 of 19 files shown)


ast/ast.go
```diff
@@ -2213,13 +2213,15 @@ type ArraySchemaType struct {
 
 // Insert is INSERT statement node.
 //
-//	INSERT INTO {{.TableName | sql}} ({{.Columns | sqlJoin ","}}) {{.Input | sql}}
+//	INSERT INTO {{.TableName | sql}} {{if .InsertOrType}}OR .InsertOrType{{end}} ({{.Columns | sqlJoin ","}}) {{.Input | sql}}
```
makenowjust, line 2216: Is this true?

I guess the following is true:

```suggestion
//	INSERT {{if .InsertOrType}}OR .InsertOrType{{end}}INTO {{.TableName | sql}} ({{.Columns | sqlJoin ","}}) {{.Input | sql}}
```

makenowjust (changes requested):

ast/ast.go
git-hulk: @makenowjust Yes, you're right.

makenowjust (approved):
LGTM! Thank you!

PR #84, opened 2024-08-15
Add support of CREATE|DROP VECTOR INDEX syntax

git-hulk:
For the syntax documentation, please refer to:

https://cloud.google.com/spanner/docs/reference/standard-sql/data-definition-language#vector_index_statements


ast/ast.go
```diff
@@ -1829,6 +1833,51 @@ type CreateIndex struct {
...
+	// pos = Create
+	// end = Options.end
+
+	Create token.Pos // position of "CREATE" keyword
+
+	IfNotExists bool // optional
+	Name        *Ident
+	TableName   *Ident
+	ColumnName  *Ident
+	Where       *Where // optional
```
makenowjust, line 1853: Is this `Where` the same as a `WHERE` clause in other statements (much like `SELECT`)?

It seems to take very restricted values, so I guess it is better to introduce a new struct instead.

makenowjust (changes requested):
I have one question. Please check it.

ast/ast.go
git-hulk: Yes, it only allows `WHERE <column_name> IS NOT NULL` for now. I reuse `WHERE` clause directly for the sake of considering it may extend more WHERE conditions in the future. Then I will restrict it to `IS NOT NULL` if you prefer this way.

makenowjust: Okay, I understand. Could you add a comment to mention that?

git-hulk: Done, can help to take a look if it's good for you. ❤️

makenowjust (approved):
Thank you!

PR #102, opened 2024-09-20
Add SplitRawStatements()

apstndb:
It is known that statement separating without parsing syntax is needed for Cloud Spanner related OSSs.

https://github.com/cloudspannerecosystem/wrench/issues/83
https://github.com/cloudspannerecosystem/yo/pull/116

memefish is good place for statement separator, because the statement separator must know about lexical structure like strings and comments 
and cloudspannerecosystem is reliable org.

Test cases should be added as
* [spanner-cli](https://github.com/cloudspannerecosystem/spanner-cli/blob/abc1cf9fced7a7d74c8344d3197f419ea5c9d95d/separator_test.go)
* [gsqlsep](https://github.com/apstndb/gsqlsep/blob/432be37d515a12c4665ca4aa963a6f4382dc750e/separator_test.go)
 
Note: I think it is no need to support custom separator like spanner-cli's `\G`.


separator_test.go
```diff
@@ -0,0 +1,49 @@
...
+)
+
+func TestSeparateRawStatements(t *testing.T) {
+	for _, test := range []struct {
+		desc  string
+		input string
+		errRe *regexp.Regexp
+		want  []string
+	}{
+		{desc: "empty input", input: "", want: []string{""}},
```
makenowjust, line 17: This spec seems inconsistent with other specs to me.

That is, `SeparateRawStatements(";")` returns `[]{""}`, so I guess that `SeparateRawStatements` omits the last empty statement and `SeparateRawStatements("")` returns `[]{}`, but this spec says it is NO.

At least, we should note this behavior in the doc.

makenowjust (commented):
The concept and naming seem good, but I have one question.

separator_test.go
apstndb: OK. I will write semantics to the go doc comment.

TL;DR

Statements are terminated by `;`, `<eof>`, or `;<eof>` . So, empty input is interpret as a single empty statement.

separator.go
```diff
@@ -0,0 +1,50 @@
...
+import "github.com/cloudspannerecosystem/memefish/token"
+
+// SeparateRawStatements separates s to statements without parsing.
+// Statements are terminated by `;`, `<eof>` or `;<eof>` and the minimum output will be []string{""}.
+// See [terminating semicolons].
+// This function won't panic but return error if lexer become error state.
+// filepath can be used in error message.
+//
+// [terminating semicolons]: https://cloud.google.com/spanner/docs/reference/standard-sql/lexical#terminating_semicolons
+func SeparateRawStatements(filepath, s string) ([]string, error) {
```
makenowjust, line 12: I think `Split` sounds better naming than `Separate`. Also, please use `src` instead of `s` because this is a public API.

makenowjust (changes requested):
I have a concern about naming.

separator.go
apstndb: > I think Split sounds better naming than Separate. 

May I better to rename to splitter.go?

> Also, please use src instead of s because this is a public API.

I believe `s string` is widely accepted in the Go community.
I prefer `input` over `src` because it's not a `src` and `dest` pattern.

makenowjust: > Would it be better to rename to splitter.go?

Yes, I think `split.go` is more preferable because this file does not define `Splitter`.

> I believe `s string` is widely accepted in the Go community.
> I prefer `input` over `src` because it's not a `src` and `dest` pattern.

`"go/parser".ParseFile` use `src`. I know this type is `any`, and more specifically it is one of `string`, `[]byte`, and `io.Reader`, but I think it is appropriate to name this argument `src`.

apstndb: > "go/parser".ParseFile use src. I know this type is any, and more specifically it is one of string, []byte, and io.Reader, but I think it is appropriate to name this argument src.

I think it is because `go/parser` is package for Go **source** files.

https://pkg.go.dev/go/parser
> Package parser implements a parser for Go source files.

I don’t feel that calling .sql files as SQL source file is majority, and this function can be used for non-file input eke stdin or `—sql` option.

makenowjust: It is major to call a string to be parsed as **source** in the parsing context, and I believe it is not different in SQL implementations. However, I have no strong opinion. Okay, I accept the argument `s`.

makenowjust: `s` sounds better than `input`.

makenowjust (approved):

makenowjust:
Thank you @apstndb!

PR #101, opened 2024-09-19
Support Spanner Graph(SQL/PGQ) schema statements

apstndb:
This PR implements Spanner Graph(SQL/PGQ) schema statements:

* `CREATE [OR REPLACE] PROPERTY GRAPH [IF NOT EXISTS]`
* `DROP PROPERTY GRAPH [IF EXISTS]`

~It conflicts with #99.~

#98

PR #111, opened 2024-09-21
Support full-text search DDLs

### LINKED ISSUE #108: Support CREATE/ALTER/DROP SEARCH INDEX
* https://cloud.google.com/spanner/docs/reference/standard-sql/data-definition-language#search_index_statements_preview
* https://cloud.google.com/spanner/docs/reference/standard-sql/data-definition-language#alter-search-index
* https://cloud.google.com/spanner/docs/reference/standard-sql/data-definition-language#drop-search-index
* `TOKENLIST` type

FYI

* `TOKENLIST` is not yet supported in spansql. https://github.com/googleapis/google-cloud-go/issues/10651

apstndb:
This PR includes `CREATE SEARCH INDEX`, `ALTER SEARCH INDEX`, and `DROP SEARCH INDEX` statements.

~Depends on helper functions in #101, so this branch includes implementation of `CREATE PROPERTY GRAPH`, and `DROP PROPERTY GRAPH`. 
It needs rebase after merge that.~ 
rebased.

* `CREATE SEARCH INDEX`
* `ALTER SEARCH INDEX`
* `DROP SEARCH INDEX`
* Ensure `CREATE TABLE` in full-text search examples can be parsed.

    ```sql
    CREATE TABLE Albums (
      AlbumId STRING(MAX) NOT NULL,
      SingerId INT64 NOT NULL,
      ReleaseTimestamp INT64 NOT NULL,
      AlbumTitle STRING(MAX),
      Rating FLOAT64,
      AlbumTitle_Tokens TOKENLIST AS (TOKENIZE_FULLTEXT(AlbumTitle)) HIDDEN,
      Rating_Tokens TOKENLIST AS (TOKENIZE_NUMBER(Rating)) HIDDEN
    ) PRIMARY KEY(AlbumId);
    ```
  * Make `STORED` optional in generated columns
    * https://cloud.google.com/spanner/docs/release-notes#June_17_2024
    > Generated columns no longer require the STORED attribute. Without this, the generated column is evaluated at query or index time and doesn't require additional storage or write overhead. For more information, see [Create and manage generated columns](https://cloud.google.com/spanner/docs/generated-column/how-to).
  * Support `HIDDEN` column

* ~Implement `ast.SearchIndexOptions` on `ast.GenericOptions` #137~
  * Need to merge after #142 is merged.
* Add `Pos.Add()` and `firstValid` to refactor `End()` https://github.com/cloudspannerecosystem/memefish/pull/111/commits/2127e0eb5495c130deb192f1ebe3d4930bd5d728

fix #108 

depends on #120


parser.go
```diff
@@ -2289,6 +2296,11 @@ func (p *Parser) parseColumnDef() *ast.ColumnDef {
 	t, notNull, null := p.parseTypeNotNull()
 	defaultExpr := p.tryParseColumnDefaultExpr()
 	generated := p.tryParseGeneratedColumnExpr()
+	hidden := p.tryExpectKeywordLike("HIDDEN")
+	hiddenPos := token.InvalidPos
+	if hidden != nil {
+		hiddenPos = hidden.Pos
+	}
```
makenowjust, line 2303:
```suggestion
	hiddenPos := token.InvalidPos
	if hidden := p.tryExpectKeywordLike("HIDDEN"); hidden != nil {
		hiddenPos = hidden.Pos
	}
```

parser.go
```diff
@@ -2405,12 +2418,18 @@ func (p *Parser) tryParseGeneratedColumnExpr() *ast.GeneratedColumnExpr {
...
 	expr := p.parseExpr()
-	p.expect(")")
-	posEnd := p.expectKeywordLike("STORED").End
+	rparen := p.expect(")").Pos
+	stored := p.tryExpectKeywordLike("STORED")
+
+	storedPos := token.InvalidPos
+	if stored != nil {
+		storedPos = stored.Pos
+	}
```
makenowjust, line 2427:
```suggestion
	storedPos := token.InvalidPos
	if 	stored := p.tryExpectKeywordLike("STORED"); stored != nil {
		storedPos = stored.Pos
	}
```

parser.go
```diff
@@ -3796,3 +3931,31 @@ func (p *Parser) errorfAtToken(tok *token.Token, msg string, params ...interface
 func (p *Parser) panicfAtToken(tok *token.Token, msg string, params ...interface{}) {
 	panic(p.errorfAtToken(tok, msg, params...))
 }
+
+// This function can't be a method because Go haven't yet supported generic methods.
+func parseSeparatedList[T interface {
+	ast.Node
+	comparable
+}](p *Parser, sep token.TokenKind, doParse func() T) []T {
```
makenowjust, line 3939: I have some concerns about this function.

- When `doParse` returns the zero value, it terminates the iteration. If the parser state is updated in `doParse`, the behavior is not intuitive.
- Is there any situation to pass a value other than `","` to `sep` in this function? Or, maybe we want to pass multiple separator tokens or more generally a function to parse separators.
- "Trailing comma" is frequently used in such situations, but is it not addressed?
- Given these issues, the function can be complex and unwieldy.
- That is, I believe that syntax separated by something is too general for parsing and should not be abstracted.

I agree with `tryExpect`. I also think that adding a function like `parseIdentList` instead of this function would not be a bad idea.

makenowjust (changes requested):
I have some minor suggestions and give a major issue. Please check it.

parser.go
apstndb: > When doParse returns the zero value, it terminates the iteration. If the parser state is updated in doParse, the behavior is not intuitive.

It may mean that treating `parse*` and `tryParse*` in the same function can be confusing.

> Is there any situation to pass a value other than "," to sep in this function? Or, maybe we want to pass multiple separator tokens or more generally a function to parse separators.

> "Trailing comma" is frequently used in such situations, but is it not addressed?

 Currently, it seems that trailing comma is [only permitted in SELECT list](https://cloud.google.com/spanner/docs/reference/standard-sql/lexical#trailing_commas) in GoogleSQL.
So, I have not yet considered to generalize it.

> Is there any situation to pass a value other than "," to sep in this function? Or, maybe we want to pass multiple separator tokens or more generally a function to parse separators.

I have considered `NEXT` of GQL.
https://cloud.google.com/spanner/docs/reference/standard-sql/graph-query-statements#gql_syntax

```
multi_linear_query_statement:
  linear_query_statement
  [
    NEXT
    linear_query_statement
  ]
  [...]
```

(It may be needed to handle `expect` and `expectKeywordLike`)
Token sequence or more general parser function as separator may be appeared, but they will be much more complex.(e.g. need to know `UNION ALL ` or `UNION DISTINCT`)

Comma separated with not `tryParse*` may be the most frequently used pattern.
Other usecases in `CREATE PROPERTY GRAPH` repo. https://github.com/cloudspannerecosystem/memefish/pull/101/files#diff-83eb8e32639d01cf443d6d8bde24c1c8be78766090d8c5f8586c36250cfedca6

```
	elements := parseSeparatedList(p, ",", p.parsePropertyGraphElement)
```

```
	list := parseSeparatedList(p, ",", func() *ast.PropertyGraphDerivedProperty {
```

apstndb: I will drop the condition of `doParse() == zero`, and research other usecases in the codebase.

makenowjust: すみません。自分のこだわり的な側面もかなりつよいので諦めて下さい。自分の目が黒いうちはこの関数がコードに入ることはないです

apstndb: OK. I will remove this abstracted function.

Instead, it is better to have the reference implementation of comma separated list to avoid bugs and code skews.

makenowjust (changes requested):

parser.go
makenowjust: Typically, parsing a comma-separated list is done by the following pattern:

```go
	xxxs := []*ast.XXX{p.parseXXX()}
	for p.Token.Kind == "," {
		p.nextToken()
		xxxs = append(xxxs, p.parseXXX())
	}
```

I quickly checked how much this idiom is used by `grep 'for p.Token.Kind == ","' parser.go | wc -l`, and it answered `19`. Therefore, we could introduce `parseCommaSepList` for that purpose. The reason I did not implement this function in the past is that Go did not have generics, so the code would not be short at all.

However, this function should always be used on a single line. That is, this usage seems wrong.

https://github.com/cloudspannerecosystem/memefish/blob/fdefffb4229337749296a510b8b525fe389e52ae/parser.go#L2664-L2672

parser.go
```diff
@@ -2595,6 +2614,107 @@ func (p *Parser) parseCreateVectorIndex(pos token.Pos) *ast.CreateVectorIndex {
...
+	p.expect("ON")
+	tableName := p.parseIdent()
+	p.expect("(")
+	columnName := parseSeparatedList(p, ",", p.parseIdent)
+	rparen := p.expect(")").Pos
+
+	storing := p.tryParseStoring()
+
+	var partitionColumns []*ast.Ident
+	if p.tryExpect("PARTITION") != nil {
```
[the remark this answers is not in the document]
apstndb: OK, `tryExpect` and `tryExpectKeywordLike` will be removed.
It may be more acceptable if the method doesn't return a value, like `func (*Parser).tryConsume()`, but it's not very useful.

parser.go
apstndb: `parseCommaSepList(p, p.parseSomething)` will be very helpful. Thanks!

makenowjust: https://github.com/cloudspannerecosystem/memefish/pull/141

apstndb:
This PR still depends on #142, so I turn this PR as draft.

apstndb:
Merge OPTIONS and conflict is resolved. Re-opened.

apstndb:
Merge to fix conflicts.

parser.go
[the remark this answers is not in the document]
apstndb: Already removed https://github.com/cloudspannerecosystem/memefish/pull/111/commits/4969ac556a2f07ef27fdb5785fd17e1ae397f122

parser.go
apstndb: Already migrated https://github.com/cloudspannerecosystem/memefish/pull/111/commits/a90745cba4b9eb574f5e1383688d7f69ea865c45

makenowjust (changes requested):

ast/ast.go
```diff
@@ -2491,6 +2513,65 @@ type ArraySchemaType struct {
...
+//
+// ================================================================================
+
+// CreateSearchIndex represents CREATE SEARCH INDEX statement
+//
+//	CREATE SEARCH INDEX {{.Name | sql}}
+//	ON {{.TableName | sql}}
+//	({{.TokenListPart | sqlJoin ", "}})
+//	{{.Storing | sqlOpt}}
+//	{{if not(.PartitionColumns | isnil)}}PARTITION BY {{.PartitionColumns  | sqlJoin ", "}}{{end}}
```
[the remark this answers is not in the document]
apstndb: You're right, I didn't know much about `text/template`.
I've found `isnil` is only needed for `StructLiteral` which is needed to distinguish `STRUCT()` and `STRUCT<>()`

makenowjust (approved):
LGTM!

PR #112, opened 2024-09-22
Add Pos, End to SplitRawStatements()

apstndb:
It changes `SplitRawStatements()` result to struct and add `Pos`, `End`.
This decision is talked in GCPUG Slack `#memefish` channel.
https://gcpug.slack.com/archives/C0518HUTECE/p1726919315091699 (This private link will be expired because Slack community plan)

It is a breaking change of exported function, but it is permitted beause memefish has no release `v0.0.0` software.

### DIFF
--- a/split.go
+++ b/split.go
@@ -2,14 +2,20 @@ package memefish
 
 import "github.com/cloudspannerecosystem/memefish/token"
 
+type RawStatement struct {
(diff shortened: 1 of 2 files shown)


makenowjust (approved):
LGTM

PR #120, opened 2024-09-29
Add helper functions for SQL(), Pos(), End()

apstndb:
Contributors need to maintain comments in `ast/ast.go` and implementations of `SQL()`, `Pos()`, and `End()` methods in `ast/sql.go` and `ast/pos.go`.
I believe productability is improved if comments and implementations can be written as similar structure.

I propose to add helper functions to implement the methods.
I believe these functions are useful as long as memefish have `ast.Node` interface .

Note
These helper functions are already used in PRs(#99, #101, #111).
`sqlOpt` needs `comparable` after Go 1.20, so this PR bump Go version to 1.20.
I have refacotored `ast.Select`, `ast.Path` and `ast.CallExpr` to demonstrate the usage and effect of helper functions.


ast/util.go
```diff
@@ -0,0 +1,91 @@
...
+			b.WriteString(sep)
+		}
+		b.WriteString(r.SQL())
+	}
+	return b.String()
+}
+
+// Helper functions for Pos(), End()
+
+// lastElem returns last element of slice s.
```
makenowjust, line 55: I understand your intention, but this function seems too generic. I would not want to add this as it could be used in places it is not supposed to be.

makenowjust (changes requested):
I have a comment to `lastElem`. Other functions are acceptable.

makenowjust:
@apstndb If these PRs are merged, do you rewrite existing implementations with them as well?

ast/util.go
apstndb: OK, `lastElem` can be written as inline, so I will remove it.

(It may be possible to be accepted if `T` is restricted to `Node`)
```diff
-func lastElem[T any](s []T) T {
+func lastElem[T Node](s []T) T {
```

apstndb:
> If these PRs are merged, do you rewrite existing implementations with them as well?

If you want to unify style of existing implementation immediately, I can rewrite them.
If not, we will rewrite them as necessary, like the Boy Scout rule.

ast/util.go
apstndb: Fixed https://github.com/cloudspannerecosystem/memefish/pull/120/commits/e1a3e815ecd3fca895317482ec3d072e999d1eb5

apstndb:
I may be better to place functions into `sql.go` and `pos.go` rather than `util.go`.

ast/util.go
makenowjust: <code>func last<del>Elem</del><ins>Node</ins>[T Node](s []T) T {</code> is acceptable.

makenowjust:
> I may be better to place functions into sql.go and pos.go rather than util.go.

This sounds good 👍

ast/util.go
apstndb: Introduce `func lastNode[T Node](s []T) T` in https://github.com/cloudspannerecosystem/memefish/pull/120/commits/47ea7094bc4ac09ecce9ef99bb7ce995f743d0e5

makenowjust (approved):
LGTM. Thank you!

Issue #100, opened 2024-09-18
Support named schema


apstndb:
I think many identities are needed to update to permit [FQNs(Full Qualified Names)](https://cloud.google.com/spanner/docs/schema-and-data-model#named-schemas).

Additionally, memefish should supports [`CREATE SCHEMA`, and `DROP SCHEMA`](https://cloud.google.com/spanner/docs/reference/standard-sql/data-definition-language#schema_statements) DDL statements.

Released at [2024-06-20](https://cloud.google.com/spanner/docs/release-notes#June_20_2024).

apstndb:
It may deprecate implicit `ast.Unnest` because it can't be distinguished from `TableExpr` with qualified table name. #123

PR #121, opened 2024-09-29
Support proto/enum as typename

apstndb:
This PR adds `NamedType`.

it is required for `PROTO`/`ENUM` support:

* [CAST AS ENUM](https://cloud.google.com/spanner/docs/reference/standard-sql/conversion_functions#cast_as_enum)
* [CAST AS PROTO](https://cloud.google.com/spanner/docs/reference/standard-sql/conversion_functions#cast_as_proto)
* [Use a protocol buffer message in a column definition](https://cloud.google.com/spanner/docs/reference/standard-sql/protocol-buffers#use_a_protocol_buffer_message_in_a_column_definition)

### DIFF
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -1396,7 +1396,7 @@ type StructField struct {
 
 // NamedType is named type node.
 // It is currently PROTO or ENUM.
-// Name is full qualified name, but it can be not to contain ".".
+// Name is full qualified name, but it can be len(Name) == 1 if it doesn't contain ".".
 //
 //	{{.Path | sqlJoin "."}}
 type NamedType struct {
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -1392,6 +1392,32 @@ type StructField struct {
 	Type  Type
 }
 
+// NamedType is named type node.
+// It is currently PROTO or ENUM.
+// Name is full qualified name, but it can be not to contain ".".
+//
+//	{{.TypeName | sql}}
+type NamedType struct {
+	// pos = Name.pos
+	// end = Name.end
+
+	Name *Path
+}
+
+func (n *NamedType) Pos() token.Pos {
+	return n.Name.Pos()
+}
+
+func (n *NamedType) End() token.Pos {
+	return n.Name.End()
+}
+
+func (n *NamedType) SQL() string {
+	return n.Name.SQL()
+}
+
+func (n NamedType) isType() {}
+
 // ================================================================================
 //
 // Cast for Special Cases
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -1404,6 +1404,8 @@ type NamedType struct {
 	Name *Path
 }
 
+func (NamedType) isSchemaType() {}
+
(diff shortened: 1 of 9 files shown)


makenowjust:
I'm concerned that it seems to be abusing `Path` for the field of `NamedType`. `Path` is a type for path expressions.

apstndb:
OK. `NamedType` will have a raw `[]*Ident`.

apstndb:
FYI
In ZetaSQL, `SELECT AS a.b.c` is parsed as `PathExpression` node.
https://github.com/google/zetasql/blob/194cd32b5d766d60e3ca442651d792c7fe54ea74/zetasql/parser/testdata/select_as_distinct_all.test#L239-L243

makenowjust:
It is abnormal that the struct with the suffix `Expr` is used as a type, ZetaSQL's design is absolutely wrong.

apstndb:
I think there is a confusion.

In GoogleSQL of Cloud Spanner(and ZetaSQL), path expression is a structure for dot-separated identifies described in lexical structure document, and it is used in many places of ZetaSQL.
https://cloud.google.com/spanner/docs/reference/standard-sql/lexical#identifier_examples
https://cloud.google.com/spanner/docs/reference/standard-sql/lexical#path_expressions

```
path:
  [path_expression][. ...]
```
(It seems they call `path` as "path expression" in many situation.)


Generally, `expr.expr` can be parsed as field access operator.
https://cloud.google.com/spanner/docs/reference/standard-sql/operators#field_access_operator.

Question: Why do we care about a path expression being an expression?

makenowjust:
A: Use a field `Path []*Ident` for that purpose like in this PR and `UpdateItem`.

https://github.com/cloudspannerecosystem/memefish/blob/8af8b4103f4d24d6516993baa17ed9eb7a768bc6/ast/ast.go#L2431

apstndb:
I think many nodes need dot-separated identifiers to implement [named schemas](https://cloud.google.com/spanner/docs/schema-and-data-model#named-schemas).
Should all they have `[]*Ident`?

apstndb:
I think it is similar to `*ast.Ident` is a `ast.Expr`, but it is used in many places.

makenowjust:
I'm sorry. I will explain the essential issue. `Path` is a list of dot-chained identifiers, where **`len(Idents) >= 2` is a constraint**. However, your use of it this time clearly violated it. Then, you needed to fix it. I feel it's my fault for not explaining that this is the biggest problem, and the GoogleSQL/ZetaSQL situation is not important here.

In the future, if ASTs similarly appear frequently where `len(Idents) > 0` (`len(Path) > 0`) is a constraint, then it is worth naming such a structure. Also, we could consider using `Path` (even if it is not an expression) if we see a dot-chained list in the syntax where `len(Idents) >= 2`.

apstndb:
Thank you for explaining.
We will discuss this topic another time. (It would be in named schema #100).

I think there are no reason to have deeper hierarchy as `NamedType{Path{[]*Ident}}`, so I agree `NamedType`  should have `[]*Ident` simply.

Issue #89, opened 2024-09-11
Support PROTO related syntax


apstndb:
https://cloud.google.com/spanner/docs/reference/standard-sql/protocol-buffers

* [x] [`[CREATE | DROP | ALTER] PROTO BUNDLE` statements](https://cloud.google.com/spanner/docs/reference/standard-sql/data-definition-language#proto_bundle_statements) https://github.com/cloudspannerecosystem/memefish/pull/176
* [x] [`CAST AS PROTO`](https://cloud.google.com/spanner/docs/reference/standard-sql/conversion_functions#cast_as_proto) #121
* [x] [`NEW` operator] https://github.com/cloudspannerecosystem/memefish/pull/183(https://cloud.google.com/spanner/docs/reference/standard-sql/operators#new_operator) #119
* [x] [`SELECT AS typename`](https://cloud.google.com/spanner/docs/reference/standard-sql/query-syntax#select_as_typename) #122
* [x] `PROTO`s  as type name #121

### FYI
These functionalities are under developed also in spansql.
* https://github.com/googleapis/google-cloud-go/issues/10944
* https://github.com/googleapis/google-cloud-go/pull/10945

apstndb:
Protocol Buffers support seems to be completed!

PR #122, opened 2024-09-29
Support SELECT AS VALUE and SELECT AS typename

### LINKED ISSUE #97: Incompatibility: SELECT AS VALUE is not supported
## Spec

https://cloud.google.com/spanner/docs/reference/standard-sql/query-syntax#select_list
```
SELECT[ compute_hint_expr ]
  [ { ALL | DISTINCT } ]
  [ AS { typename | STRUCT | VALUE] } ]
  select_list
```

https://cloud.google.com/spanner/docs/reference/standard-sql/query-syntax#select_as_value
https://cloud.google.com/spanner/docs/reference/standard-sql/query-syntax#value_tables

## Reproducing input

```
SELECT * FROM (SELECT AS VALUE STRUCT(123 AS a, FALSE AS b))
```

## Current error

```
syntax error::1:26: expected token: STRUCT, but: <ident>

  1:  SELECT * FROM (SELECT AS VALUE STRUCT(123 AS a, FALSE AS b))
                               ^~~~~
```

apstndb:
This PR add support of `SELECT AS VALUE` #97  and `SELECT AS typename` #89.

`SELECT AS VALUE`
https://cloud.google.com/spanner/docs/reference/standard-sql/query-syntax#select_as_value

`SELECT AS typename`
https://cloud.google.com/spanner/docs/reference/standard-sql/query-syntax#select_as_typename

They and `SELECT AS STRUCT` are generalized as `ast.SelectAs`.

~Note: This PR depends on `ast.NamedType` in #121, so it is needed to rebase after #121 is merged.~ rebased

fix #97

### DIFF
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -428,7 +428,7 @@ type CTE struct {
 //
 //	SELECT
 //	  {{if .Distinct}}DISTINCT{{end}}
-//	  {{if .AsStruct}}AS STRUCT{{end}}
+//	  {{.As | sqlOpt}}
 //	  {{.Results | sqlJoin ","}}
 //	  {{.From | sqlOpt}}
 //	  {{.Where | sqlOpt}}
@@ -443,7 +443,8 @@ type Select struct {
 	Select token.Pos // position of "select" keyword
 
 	Distinct bool
-	AsStruct bool
+	AsStruct bool // deprecated, use As
+	As       SelectAs
 	Results  []SelectItem // len(Results) > 0
 	From     *From        // optional
 	Where    *Where       // optional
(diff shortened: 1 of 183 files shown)


apstndb:
Rebase to reduce dependency to other branch.

ast/ast.go
```diff
@@ -443,7 +453,8 @@ type Select struct {
 	Select token.Pos // position of "select" keyword
 
 	Distinct bool
-	AsStruct bool
+	AsStruct bool         // deprecated, use As
```
makenowjust, line 456: I'm not sure this deprecation is necessary. Could we remove it simply?

apstndb: Removed https://github.com/cloudspannerecosystem/memefish/pull/122/commits/ff14fe6ee2485c523010146f06cc6ff60b8d37a8

It is because I didn't sure the deprecation policy of memefish.
Current policy: We can change any AST node without breaking change notification because there are no release version.
Is it correct?

makenowjust: I have no time to maintain codes with some deprecations. Also, we have no plan to publish a major version, so we can easily introduce breaking changes IMO.

apstndb: Thank you for answering!

We can make breaking changes to v0.x.y at any time, but the release process itself is a significant effort, so I think the current policy is fine.

PR #142, opened 2024-10-18
Generalize OPTIONS

### LINKED ISSUE #137: Generalize OPTIONS behavior
Currently, memefish fails unknown option key in some `OPTIONS`.
`OPTIONS` key can be added without notification, and it is even possible to have undocument options.

I think it is better to have "schema on read" semantics rather than "schema on write(parse)" semantics.
Validation should be the responsibility of semantic analysis phase.

Proposed characteristics

* Any key-value pairs can be parsed and unparsed.
* Known key-value can be accessed by named methods.
  * It can be error if invalid value type, and the error can be handled.
  * These are maintained on a best effort basis.
* Concrete options share underlying implementation as `GenericOptions`.

apstndb:
This PR generalize all options implementation to `Options`.

Note:
* I named  for single element of `Options` as `OptionsDef`. It is because `options_def` is used in [Spanner GoogleSQL DDL reference](https://cloud.google.com/spanner/docs/reference/standard-sql/data-definition-language).
* `OptionsDef.SQL()` prints `null`, `true`, `false` as lowercase because it is consistently used in docs.
* While it is inconsistent in official docs,`OptionsDef.SQL()` prints whitespaces around `=`. 

fixes #137


ast/ast.go
```diff
@@ -1445,6 +1448,115 @@ type CastNumValue struct {
...
+
+	Options token.Pos // position of "OPTIONS" keyword
+	Rparen  token.Pos // position of ")"
+
+	Records []*OptionsDef // len(Records) > 0
+}
+
+// Field finds name in Records, and return its value as Expr.
+// The second return value indicates that the name was found in Records.
+// It is not exposed because you can use *Field methods.
```
makenowjust, line 1466: Is it true?

makenowjust (changes requested):
Could you move `Options` methods into a new file (e.g., `options.go`)?

ast/ast.go
apstndb: Oh, I forgot to fix comment.
It is not true. I will fix it.

apstndb: Fixed https://github.com/cloudspannerecosystem/memefish/pull/142/commits/b469bb938f7e51284a572b45f0895f909079c79d

ast/ast.go
```diff
@@ -1445,6 +1445,29 @@ type CastNumValue struct {
...
+type Options struct {
+	// pos = Options
+	// end = Rparen + 1
+
+	Options token.Pos // position of "OPTIONS" keyword
+	Rparen  token.Pos // position of ")"
+
+	Records []*OptionsDef // len(Records) > 0
+}
+// OptionsDef is single option definition for DDL statements.
```
makenowjust, line 1460: Insert a newline here.

```suggestion
}

// OptionsDef is single option definition for DDL statements.
```

ast/ast.go
```diff
@@ -1497,7 +1520,7 @@ type CreateSequence struct {
 
 	Name        *Ident
 	IfNotExists bool
-	Options     *SequenceOptions
+	Options     *Options
```
makenowjust, line 1523:
```suggestion
	Options     *Options // optional
```

Also, we should update the document comment for this struct.

makenowjust (changes requested):

parser.go
```diff
@@ -2522,28 +2448,34 @@ func (p *Parser) parseOnDeleteAction() (onDelete ast.OnDeleteAction, onDeleteEnd
...
+}
+
+func (p *Parser) parseOptions() *ast.Options {
+	optionsPos := p.expectKeywordLike("OPTIONS").Pos
+	p.expect("(")
+	optionsDefs := parseCommaSeparatedList(p, p.parseOptionsDef)
+	rparen := p.expect(")").Pos
+
+	return &ast.Options{
+		Options: optionsPos,
```
makenowjust, line 2475: `pos` is preferred here.

```suggestion
	pos := p.expectKeywordLike("OPTIONS").Pos
	p.expect("(")
	optionsDefs := parseCommaSeparatedList(p, p.parseOptionsDef)
	rparen := p.expect(")").Pos

	return &ast.Options{
		Options: pos,
```

apstndb:
Thank you for review, all suggestions are applied!

ast/ast.go
makenowjust: Repeat:

> we should update the document comment for this struct.

apstndb: I have noticed that the current implementation still treat `OPTIONS` clause as required.
https://github.com/apstndb/memefish/blob/1e59f4158f2626d063128898c42fbd804f6c60b8/parser.go#L2183

It is because there is required option `sequence_kind`, so `OPTIONS` is not optional.
> The `sequence_kind` option accepts a `STRING` to indicate the type of sequence to use. At this time, `bit_reversed_positive` is the only valid type and it's a required option.

```
spanner> CREATE SEQUENCE MySequence;
ERROR: rpc error: code = InvalidArgument desc = MySequence does not have a valid sequence kind. Please specify the sequence kind explicitly or set the database option `default_sequence_kind`.
```

We can parse `OPTIONS` as optional to ready future changes, should it be?

makenowjust: Ok, I understand that `Options` should be required here, so let's revert it.

But, repeat,

> we should update the document comment for this struct.

Currently,

```go
//	CREATE SEQUENCE {{if .IfNotExists}}IF NOT EXISTS{{end}} {{.Name | sql}} }} OPTIONS ({{.Options | sqlJoin ","}})
```

but the following is corret I think:

```go
//	CREATE SEQUENCE {{if .IfNotExists}}IF NOT EXISTS{{end}} {{.Name | sql}} {{.Options | sql}}
```

apstndb: Thank you!
I believe it is correct changes https://github.com/cloudspannerecosystem/memefish/pull/142/commits/405a7c5bdcc5bd676116697d6522642dfb5326a4.

makenowjust (approved):
Good!

PR #149, opened 2024-10-19
Fix some `Pos` and `End` issues on `TableConstraint`

makenowjust:
- In `TableConstraint` definition, comments about `pos` and `end` were missing. This PR adds them.
- When `ConstraintPos` is missing, it sets `InvalidPos` explicitly.
- Also, that `ForeignKey` and `Check` set `expect(")").End` to `Rparen` on parsing seems incorrect, and `expect(")").Pos` is correct.

Note that `expect(")").End` also appears in `parseColumnDefOptions`, but it will be removed in #142.


apstndb (approved):
I am not a reviewer role, but LGTM.

makenowjust:
Thank you @apstndb.

PR #158, opened 2024-10-21
Support named schema in existing DDL statements

apstndb:
This PR implements named schema support in existing DDL statements. #100 

* I believe `testdata/input/ddl/named_schema_*.sql` covers current [named schema](https://cloud.google.com/spanner/docs/schema-and-data-model#named-schemas) use cases.
* Many breaking changes because of replacing `*Ident` with `*Path`.
  * It need to relax the current constraints of `Path.Idents` to permit `len(Idents) == 1`

Note

~I am thinking That using `Path` as FQN implementation is natural because is is possible to be extended to multi level named schema, but temporarily implemented as FQN{Name string; Schema string} for PoC.~


apstndb:
Turn to draft because it is not completed.

makenowjust (approved):
LGTM. Thank you!

PR #160, opened 2024-10-21
Sort `Statement` and `DDL` types

makenowjust:
This order follows the official documentation:

- https://cloud.google.com/spanner/docs/reference/standard-sql/data-definition-language
- https://cloud.google.com/spanner/docs/reference/standard-sql/dml-syntax

If this rule can be enforced by a linter, it sounds really good.


ast/ast.go
```diff
@@ -53,28 +53,28 @@ type Statement interface {
...
+func (CreateView) isStatement()         {}
+func (CreateChangeStream) isStatement() {}
 func (CreateRole) isStatement()         {}
+func (CreateSequence) isStatement()     {}
+func (CreateVectorIndex) isStatement()  {}
 func (AlterTable) isStatement()         {}
 func (AlterIndex) isStatement()         {}
+func (AlterChangeStream) isStatement()  {}
 func (AlterSequence) isStatement()      {}
 func (DropTable) isStatement()          {}
```
apstndb, line 66: Comment in PR says

> This order follows the official documentation:

but this ordering sorts `Create`, `Alter`, `Drop` first, and object type second.

Official document ordering is object type first, `Create`, `Alter`, `Drop` second.

makenowjust: After one night, I think it is better to use the official doc order directly because we can quickly check which statements are missing in memefish by comparing the official doc list.

@apstndb What do you think?

apstndb: I agree that the official doc order itself is better ordering because:

* I feel official documented order is natural for contributors.
* There are many DDLs without `CREATE`, `ALTER`, `DROP` prefix so I feel manipulation-based ordering is hard to define consistently.
* Usually, DDLs for the same object type are implemented in the same PR(or consective PRs), so diff will be more cunsecutive in object-based ordering.

ast/ast_test.go
```diff
@@ -8,21 +8,28 @@ func TestStatement(t *testing.T) {
...
 	Statement(&CreateDatabase{}).isStatement()
 	Statement(&CreateTable{}).isStatement()
-	Statement(&CreateView{}).isStatement()
-	Statement(&CreateIndex{}).isStatement()
-	Statement(&CreateSequence{}).isStatement()
-	Statement(&CreateRole{}).isStatement()
 	Statement(&AlterTable{}).isStatement()
-	Statement(&AlterIndex{}).isStatement()
 	Statement(&DropTable{}).isStatement()
+	Statement(&CreateIndex{}).isStatement()
```
apstndb, line 13: Is comments not needed in `ast_test.go`?

apstndb (approved):
LGTM for ordering.
If the sort order is needed to be enforced, I think it should be commented.

ast/ast_test.go
makenowjust: No, for now. I'd like to remove `ast_test.go`. This is invaluable. What are we testing with this?

apstndb: I agree to remove.
Most interface implementation are already tested in `parse*()` in `parser.go`, and maintainance ordering in `ast_test.go` is waste of time.
May only valuable thing be missing `TestNode()` because it is possible to fail to implement methods of `Node`, and it can't be detected unless it is used as `Node`.

Issue #94, opened 2024-09-12
Incompatibility: SAFE functions are not supported


apstndb:
This incompatibility is detected by #90.

- [ ] `SAFE.` prefix
- [x] `SAFE_CAST` #164
- [x] `SAFE_ORDINAL`, `SAFE_OFFSET` #192 

## Spec

https://cloud.google.com/spanner/docs/reference/standard-sql/functions-reference#safe_prefix

Additionally, [`SAFE_CAST`](https://cloud.google.com/spanner/docs/reference/standard-sql/conversion_functions#safe_casting) is needed to be supported. #115 

## Reproducing input
```
CREATE TABLE T (
  id INT64,
  string_field STRING(MAX),
  gen BYTES(MAX) AS (SAFE.MD5(string_field)) STORED,
) PRIMARY KEY(id);
```

## Current error
```
syntax error:testdata/generated_column.test:5:30: expected token: ), but: (

  5:    gen BYTES(MAX) AS (SAFE.MD5(string_field)) STORED,
                                   ^
```

PR #164, opened 2024-10-21
Implement SAFE_CAST in CastExpr

apstndb:
This PR implements `SAFE_CAST` as `Safe` flag in `CastExpr`.

part of #94


makenowjust (approved):
LGTM

PR #167, opened 2024-10-22
Follow the naming convention for error values

makenowjust:
In many Go libraries, names of error values are prefixed with `Err`. We also follow this convention.

### DIFF
--- a/ast/options.go
+++ b/ast/options.go
@@ -23,11 +23,11 @@ func (o *Options) Field(name string) (expr Expr, found bool) {
 
 // FieldNotFound is returned Options.*Field methods.
 // It can be handled as a non-error.
-var FieldNotFound = errors.New("field not found")
+var ErrFieldNotFound = errors.New("field not found")
 
 // fieldTypeMismatch is only defined for test.
 // It is intentionally unexported.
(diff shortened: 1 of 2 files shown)


ast/options.go
```diff
@@ -23,11 +23,11 @@ func (o *Options) Field(name string) (expr Expr, found bool) {
 
 // FieldNotFound is returned Options.*Field methods.
 // It can be handled as a non-error.
-var FieldNotFound = errors.New("field not found")
+var ErrFieldNotFound = errors.New("field not found")
 
 // fieldTypeMismatch is only defined for test.
 // It is intentionally unexported.
-var fieldTypeMismatch = errors.New("type mismatched")
+var errFieldTypeMismatch = errors.New("type mismatched")
```
makenowjust, line 30: @apstndb A question unrelated to this issue: Why is this value not exported?

apstndb: It is because user should only check `ErrFieldNotFound` which is not error state. Other errors will be handled as unrecoverable error.
If user want to handle unexpected types, `(*Options).Field(name string) (Expr, bool)` should be used.

apstndb (approved):
Indeed LGTM.

Issue #151, opened 2024-10-19
Refactor SQL(), Pos(), End()


apstndb:
After introducing helper functions of `SQL()`, `Pos()`, `End()`, there are mixed style implementation of their functions.

It is better to refactoring existing functions, but they are not yet done.
https://github.com/cloudspannerecosystem/memefish/pull/120#issuecomment-2412866760

- [x] Rewrite `SQL()` #173 
- [x] ~Rewrite `Pos(), End()`~
  - It is resolved by #174

apstndb:
Close as completed

PR #173, opened 2024-10-23
Refactor SQL() methods

apstndb:
This PR refactors `SQL()` methods.

* This PR only modifies `ast/sql.go`
* The correctness of the refactoring is ensured by not making any modifications to `testdata`.
* Some of the current implementation is incorrect, so they will be fixed in a separate PR.
  * ~Empty struct is better to be unparse as `STRUCT()`, but currently `STRUCT<>()`.~ discussed https://github.com/cloudspannerecosystem/memefish/pull/173#pullrequestreview-2391072052
  * `ast.ChangeStreamForTable` doesn't implement `ast.Node`
* It is a part of #151 

Note:
There is a idea to introduce `text/template` style conditions.

https://pkg.go.dev/text/template#hdr-Actions

```
{{if pipeline}} T1 {{end}}
	If the value of the pipeline is empty, no output is generated;
	otherwise, T1 is executed. The empty values are false, 0, any
	nil pointer or interface value, and any array, slice, map, or
	string of length zero.
	Dot is unaffected.
```

It can simplify more, but empty slice, map handling may require to use `reflect` because empty slice/map are not the same with zero values, and their types doesn't satisfy `comparable`.

Added: I have implement only zero value as false, but it seems not useful than expected: https://github.com/cloudspannerecosystem/memefish/commit/1606256407063cdb89668b3ca6cb304593065cd6


ast/sql.go
```diff
@@ -621,38 +491,18 @@ func (a *ArrayLiteral) SQL() string {
...
-	sql += "("
-	for i, v := range s.Values {
-		if i != 0 {
-			sql += ", "
-		}
-		sql += v.SQL()
-	}
-	sql += ")"
-	return sql
+	// TODO: len(s.Fields) > 0 is better than s.Fields != nil, but it need to update current testdata.
```
makenowjust, line 494: We intentionally distinct `STRUCT` and `STRUCT<>` because the first means a struct having any fields, and the second is the struct with no fields. For instance, `select CAST(struct(1) AS STRUCT<x INT64>).x` is valid, but `select CAST(struct<>(1) AS STRUCT<x INT64>).x` is invalid.

apstndb: Thank you for your explanation.
I understood your intention.

> We intentionally distinct STRUCT and STRUCT<> because the first means a struct having any fields, and the second is the struct with no fields. For instance, select `CAST(struct(1) AS STRUCT<x INT64>).x` is valid, but select `CAST(struct<>(1) AS STRUCT<x INT64>).x` is invalid.

My thought is that:
* It is `StructLiteral`, not `StructType`, so there is no semantic difference between `STRUCT<>()` and `STRUCT()`. It can be normalized without ambiguity.
* `STRUCT<>()` is valid, but it can feel unfamiliar because it has no fields to name or type.
* The official docs describe only [`STRUCT()`](https://cloud.google.com/spanner/docs/reference/standard-sql/data-types#typeless_struct_syntax).

It might be a good idea to comment that both `STRUCT()` and `STRUCT<>()` are lexically preserved.

makenowjust: Perhaps, we need to split the current `StructLiteral` into a new two s.t. `TypedStructLiteral` and `TypelessStructLiteral` for clarity.

apstndb: LGTM for splitting.

In ZetaSQL, both of them are treated in [`ASTStructConstructorWithKeyword`](https://github.com/google/zetasql/blob/194cd32b5d766d60e3ca442651d792c7fe54ea74/zetasql/parser/gen_parse_tree.py#L2757), but they can't be mixed.
* In [typed struct syntax](https://cloud.google.com/spanner/docs/reference/standard-sql/data-types#struct_type)
    > `AS alias` is not allowed on the input expressions.
     > STRUCT<x int64>(5 AS x) | Error - Typed syntax does not allow AS
     > -- | --

apstndb:
Thought: The `SQL()` definition does not use `token.Pos.Invalid()` without negation, so why is there no `token.Pos.Valid()`?

makenowjust (approved):
Good. Thank you.

PR #174, opened 2024-10-24
Generate `ast/pos.go` automatically

makenowjust:
This PR introduces the parser for the POS expressions (the pos/end notation in `ast/ast.go`) and the generator for `ast/pos.go` using it.
`Makefile` is also updated, so `go generate` is executed automatically in most cases.

Also, this PR adds `-dig` and `-pos` options to `tools/parser` for testing the POS expression.


apstndb:
That's a great thing for us contributors.
We can now focus on getting the `Node` definition right without having to worry about `pos.go`.

If you merge it, it will resolve the conflict in my PR.

Since the importance of `tools/parse` in the development flow seems to have increased, it would be nice to have examples documented on how to use `-dig` and `-pos`.

makenowjust:
@apstndb Some examples are added to the usage of `tool/parse/main.go`.

PR #179, opened 2024-10-27
Support SCHEMA DDL statements

apstndb:
This PR implements `SCHEMA` DDL statements, `CREATE SCHEMA` and `DROP SCHEMA`.

Note

The official docs says `DROP SCHEMA` can target multiple schema names.

```
DROP SCHEMA schema_name [, ...]
```

but, it seems not to be true.

```
spanner> DROP SCHEMA schema1, schema2;
ERROR: rpc error: code = InvalidArgument desc = Error parsing Spanner DDL statement: DROP SCHEMA schema1, schema2 : Syntax error on line 1, column 20: Expecting 'EOF' but found ','
```

* part of #100


makenowjust (approved):
LGTM

PR #119, opened 2024-09-28
Support protobuf NEW operators

apstndb:
Part of #89 

This PR adds `NEW` operators for Protocol Buffers.

https://cloud.google.com/spanner/docs/reference/standard-sql/operators#new_operator

> `NEW protocol_buffer {...}`: Creates a protocol buffer using a map constructor.

```
NEW protocol_buffer {
  field_name: literal_or_expression
  field_name { ... }
  repeated_field_name: [literal_or_expression, ... ]
}
```
> `NEW protocol_buffer (...)`: Creates a protocol buffer using a parenthesized list of arguments.

```
NEW protocol_buffer(field [AS alias], ...field [AS alias])
```

PR #183, opened 2024-10-27
Introduce `TypelessStructLiteralArg` and `NewConstructorArg` interfaces

makenowjust:
Close #119 
Close #177

~`TypelessStructValue` (#177) and `NewConstructorValue` (#119) share their struct shapes. Then, this PR introduces a new struct named `AsExpr` for `expression AS name`-kind values and replaces them with it.~

~I believe this is not a hasty generalization because these struct shapes are exactly the same, and they are used for common semantics. However, I'm not sure `AsExpr` is the best name. If you have a better name, please tell me it!~

~Thank you.~

- - -

See https://github.com/cloudspannerecosystem/memefish/pull/183#issuecomment-2441417390.


apstndb:
> `TypelessStructValue` (#177) and `NewConstructorValue` (#119) share their struct shapes. Then, this PR introduces a new struct named `AsExpr` for `expression AS name`-kind values and replaces them with it.
> 
> I believe this is not a hasty generalization because these struct shapes are exactly the same, and they are used for common semantics.

I have concerned about there are undocumented differences in `NEW` constructors and typeless `STRUCT` literals.

https://github.com/google/zetasql/blob/194cd32b5d766d60e3ca442651d792c7fe54ea74/zetasql/parser/bison_parser.y#L8927-L8942

```
    | expression "AS" "(" path_expression ")"
      {
        // Do not parenthesize $4 because it is not really a parenthesized
        // path expression. The parentheses are just part of the syntax here.
        $$ = MAKE_NODE(ASTNewConstructorArg, @$, {$1, $4});
      }
```

I believe it is a syntax for proto extensions, and it is not yet implemented in Cloud Spanner, but I think these semantics are not same. It may be a reason to distinguish between `TypelessStructValue ` and `NewConstructorValue `.

```
spanner> SELECT NEW examples.shipping.`Order`("Japan" AS (shipping_address.country));
ERROR: spanner: code = "InvalidArgument", desc = "NEW constructor does not support proto extensions [at 1:50]\\nSELECT NEW examples.shipping.`Order`(\\\"Japan\\\" AS (shipping_address.country))\\n      
```

apstndb:
* `NewConstructorValue ` is a structure for `PROTO` type.
* `TypelessStructLiteral` is a structure for `STRUCT` type.

apstndb:
Personally I think they can be generalized temporarily, even if their semantics are not same,
But I am wondering if this generalization really have benefit.

> we can introduce a new interface NewConstructorArg, and a new struct `ExprAsPath`

I think it is not possible because full definition is:

```
new_constructor_arg:
    expression
      {
        $$ = MAKE_NODE(ASTNewConstructorArg, @$, {$1});
      }
    | expression "AS" identifier
      {
        $$ = MAKE_NODE(ASTNewConstructorArg, @$, {$1, $3});
      }
    | expression "AS" "(" path_expression ")"
      {
        // Do not parenthesize $4 because it is not really a parenthesized
        // path expression. The parentheses are just part of the syntax here.
        $$ = MAKE_NODE(ASTNewConstructorArg, @$, {$1, $4});
      }
    ;
```
It is `expression AS (path_expression)`, not `expression AS path_expression`.
`NewConstructorArg` is special structure with no other similar syntax, because it is specialized to Protocol Buffers.

apstndb:
I understood. We can add another interface if it is needed.

`NewConstructorArg` won't be concrete struct.

apstndb:
> However, I'm not sure `AsExpr` is the best name. If you have a better name, please tell me it!
> [Rename AsExpr to ExprAsName, and use AsAlias](https://github.com/cloudspannerecosystem/memefish/pull/183/commits/449b666950daf47502f720dd95217e76258c29aa)

IMHO, `ExprAsName` doesn't imply the name is optional, and there are other `expr AS name` structure(e.g. in `* REPLACE()`).
So more explicit name, like `ExprOptName`, may be unambiguous.

parser.go
```diff
@@ -1798,16 +1807,15 @@ func (p *Parser) parseParenExpr() ast.Expr {
 	}
 
 	if p.Token.Kind != "," {
-		p.panicfAtToken(&paren, "cannot parse (...) as expression, struct literal or subquery")
+		p.panicfAtToken(&paren, "cannot parse (...) as expression, typeless struct literal or subquery")
```
apstndb, line 1810:
```suggestion
		p.panicfAtToken(&paren, "cannot parse (...) as expression, tuple struct literal or subquery")
```

ast/ast.go
```diff
@@ -1376,22 +1377,56 @@ type ArrayLiteral struct {
...
+	// pos = Struct
+	// end = Rparen + 1
+
+	Struct token.Pos // position of "STRUCT"
+	Rparen token.Pos // position of ")"
+
+	Values []*ExprAsName
+}
+
+// ExprAsName is value with optional name in typeless struct literal or new constructor.
```
apstndb, line 1419: I think it is not only for typeless struct literal and new constructor.
We can encourage to use this struct for any `expr [AS ident]` syntax, without any semantic implications.

makenowjust: We can update the document anytime.

apstndb: It means that this struct is added only for `NewConstructorArg` and `TypelessStructValue` in this PR context and we need to discuss to change this struct semantics in another PR.

makenowjust: Yes.

apstndb: From this conversation, it sounds like you don't intend to widely recommend the use of this struct.

apstndb: My thought:
* `TypelessStructValue` and `NewConstructorArg` is not same abstraction, but have common concrete structure.
* Concrete structure will able to be reused without modification.
* IMO, currently `ExprAsName` is no strong reason to be introduced.

makenowjust:
Finally, I conclude there is no better name for the struct `expression [AS name]`. That is, we should not introduce such a struct.

Therefore, we introduce new interfaces `TypelessStructLiteralArg` and `NewConstructorArg`. Currently, two interfaces have the same implementations and parsers, but they are separated for future extensibility.

parser.go
```diff
@@ -479,22 +479,26 @@ func (p *Parser) parseSelectItem() ast.SelectItem {
 	}
 }
 
-func (p *Parser) tryParseAsAlias() *ast.AsAlias {
+func (p *Parser) tryParseAsAlias(requiredAs bool) *ast.AsAlias {
```
apstndb, line 482: I feel like `p.tryParseAsAlias( /* requiredAs */ false)` is a little difficult to write.
Can't we introduce two method for `requiredAs=true` and `requiredAs=False`? (like (`tryParseAsAliasAsRequired` and `tryParseAsAliasAsOptional`)

apstndb (commented):
I feel it is natural to introduce these interfaces and use `ExprArg` and `Alias`.
I have not noticed their `SelectItem` implementation can be used to implement them.

I agree their implementations are better than my original PRs.

parser.go
makenowjust: `tryParseAsAliasAsRequired` seems like a strange name to someone unfamiliar with this project (e.g. twice `As` looks like typo, and `try` and `required` seem contradictory), and it is hard to guess what the method does. By making the explicit argument, it is clear that the method parses `AS name` as possible, and the comment makes it clear that the argument gives the required `AS`.

apstndb: My thought:
> twice As looks like typo and try and required seem contradictory
I feel that the current `AsAlias` can be `Alias`, and the current `Alias` is like `ExprWithAlias` or `AliasedExpr`.
`tryParseAliasAsRequired`, and `tryParseAliasAsOptional` is felt better.

> the comment makes it clear that the argument gives the required AS.

I think maintaining `/* requiredAs */` comment is harder because it can't be assisted by go-lsp or IDEs, and it is strange to someone(me) unfamiliar with this project. At least, it must be noted in doc comment of `tryParseAsAlias()`.

apstndb: > `tryParseAliasAsRequired` also seems strange to me because `As` looks like a usual preposition, and `try` and `required` have conflicting meanings.

It means it can be better if the subject of "required" or "optional" is unambiguous.
The subject is "keyword `AS`", so they can be called as `tryParseAsAliasRequireKeyword`, `tryParseAsAliasOptionalKeyword`.

makenowjust: It's ambiguous enough that there are multiple methods to parse a particular struct, and I don't think it's a good idea to distinguish them by method names. If it really should be parsed by multiple methods, we should introduce multiple structs, but I don't think in this case.

apstndb: > It's ambiguous enough that there are multiple methods to parse a particular struct

I agree structs and methods should be one by one.
In the other language, it may be a usecase of missing named parameter.
In Go, it may be better to introduce const with enum semantics.

makenowjust: Okay, because of code jumps and editor popups, I decided to remove the `/* requiredAs */` comment.

apstndb: We have agreed about `tryParseAsAlias()`, so it is just my thought.
> I think that `Alias` is the identifier after `AS`, not the `AS name` itself. 

SQL has common structure "alias, which is optionally prefixed by `AS`", and they are safe to unparse always with `AS`.
I prefer to call it `Alias` than `AsAlias` because someones have coding guideline which prefer to omit `AS`.
As far as I know, `AS` of table alias is often omitted. (Probably due to Oracle's past not allowing `AS`.)

makenowjust: The name `Alias` ​​has less information than the name `AsAlias`, so I don't think it should be changed proactively.

apstndb (approved):
I believe we have finished discussions and `tryParseAsAlias` with typed enum resolves our concernings.
Thank you for finishing my PRs.
LGTM!

PR #187, opened 2024-10-30
Fix error positions on tokenizing a literal terminated with a newline

makenowjust:
Before:

```console
go run ./tools/parse/main.go $'"foo\n'
syntax error: :1:1: unclosed string literal: newline appears in non triple-quoted

  1:  "foo
  2:
exit status 1
```

After:

```console
$ go run ./tools/parse/main.go $'"foo\n'
syntax error: :1:1: unclosed string literal: newline appears in non triple-quoted

  1:  "foo
      ^~~~
exit status 1
```


apstndb (approved):
It is the behavior I expected. LTGM!

PR #188, opened 2024-10-30
Improve error messages on parsing POS of `gen-ast-pos`

makenowjust:
This PR fixes error message on parsing POS expressions to show a node name and an expression. I believe this fix helps our development.


apstndb (approved):
It is what I needed, LGTM!

PR #189, opened 2024-10-31
Introduce `BadNode` and `ErrorList`

makenowjust:
This PR introduces `BadNode` and `ErrorList`.

- `BadNode` is a placeholder node for a source fragment containing syntax errors.
- `ErrorList` is an error consisting of a list of `*memefish.Error`.
- Now, the parser can report multiple errors and return an AST node even if errors are reported.

Example:

```console
$ go run ./tools/parse/main.go 'select (1 +) * (% 2)'
--- Error
syntax error: :1:12: unexpected token: )

  1:  select (1 +) * (% 2)
                 ^


syntax error: :1:17: unexpected token: %

  1:  select (1 +) * (% 2)
                      ^


--- AST
&ast.QueryStatement{
  Hint:  (*ast.Hint)(nil),
  With:  (*ast.With)(nil),
  Query: &ast.Select{
    Select:   0,
    Distinct: false,
    As:       nil,
    Results:  []ast.SelectItem{
      &ast.ExprSelectItem{
        Expr: &ast.BinaryExpr{
          Op:   "*",
          Left: &ast.ParenExpr{
            Lparen: 7,
            Rparen: 11,
            Expr:   &ast.BadNode{
              NodePos: 8,
              NodeEnd: 11,
              Raw:     "1 +",
            },
          },
          Right: &ast.ParenExpr{
            Lparen: 15,
            Rparen: 19,
            Expr:   &ast.BadNode{
              NodePos: 16,
              NodeEnd: 19,
              Raw:     "% 2",
            },
          },
        },
      },
    },
    From:    (*ast.From)(nil),
    Where:   (*ast.Where)(nil),
    GroupBy: (*ast.GroupBy)(nil),
    Having:  (*ast.Having)(nil),
    OrderBy: (*ast.OrderBy)(nil),
    Limit:   (*ast.Limit)(nil),
  },
}

--- SQL
SELECT (1 +) * (% 2)
```


parser.go
```diff
@@ -2293,85 +2300,121 @@ func (p *Parser) lookaheadSimpleType() bool {
...
+
 	pos := p.Token.Pos
 	switch {
 	case p.Token.Kind == "CREATE":
 		p.nextToken()
 		switch {
 		case p.Token.IsKeywordLike("SCHEMA"):
-			return p.parseCreateSchema(pos)
+			ddl = p.parseCreateSchema(pos)
+			return
```
apstndb, line 2318: I think explicit return is preferred than naked returns even if you use named result parameters.

https://go.dev/wiki/CodeReviewComments#named-result-parameters
> Naked returns are okay if the function is a handful of lines. Once it’s a medium sized function, be explicit with your return values.

PR #196, opened 2024-11-08
Implement EXCEPT and REPLACE in select item

### LINKED ISSUE #184: Support EXCEPT and REPLACE in SelectItem
```
select_all:
  [ expression. ]*
  [ EXCEPT ( column_name [, ...] ) ]
  [ REPLACE ( expression [ AS ] column_name [, ...] ) ]
```

## EXCEPT

https://cloud.google.com/spanner/docs/reference/standard-sql/query-syntax#select_except

```
$ go run ./tools/parse --mode query "SELECT * EXCEPT (n) FROM (SELECT 1 AS n, "foo" AS s)"                    
2024/10/28 00:12:11 syntax error: :1:17: expected token: ALL, DISTINCT, but: (

  1:  SELECT * EXCEPT (n) FROM (SELECT 1 AS n, foo AS s)
```

## REPLACE

https://cloud.google.com/spanner/docs/reference/standard-sql/query-syntax#select_replace

```
$ go run ./tools/parse --mode query "SELECT * REPLACE (2 AS n) FROM (SELECT 1 AS n, "foo" AS s)"
2024/10/28 00:12:44 syntax error: :1:10: expected token: <eof>, but: <ident>

  1:  SELECT * REPLACE (2 AS n) FROM (SELECT 1 AS n, foo AS s)
               ^~~~~~~
```

Note:

Official syntax docs says `AS` is optional, but actually `AS` can't be omitted

```
spanner> SELECT * REPLACE (2 n) FROM (SELECT 1 AS n, "foo" AS s);
ERROR: spanner: code = "InvalidArgument", desc = "Syntax error: Expected keyword AS but got identifier \\\"n\\\" [at 1:21]\\nSELECT * REPLACE (2 n) FROM (SELECT 1 AS n, \\\"foo\\\" AS s)\\n                    ^"
```

apstndb:
This PR implements `EXCEPT` and `REPLACE` in `SelectItem`.

Note:
* There is no breaking changes.
* Introduced `p.lookaheadStarModifierExcept()` to avoid ambiguity between `* EXCEPT (columns)` and `* EXCEPT ALL` or `* EXCEPT DISTINCT`.
* The prefix `StarModifier` is borrowed from ZetaSQL:  https://github.com/google/zetasql/blob/194cd32b5d766d60e3ca442651d792c7fe54ea74/zetasql/parser/gen_parse_tree.py#L3617-L3634

fixes #184


ast/ast.go
```diff
@@ -620,26 +620,71 @@ type SubQuery struct {
...
+//
+//	REPLACE ({{Columns | sqlJoin ", "}})
+type StarModifierReplace struct {
+	// pos = Replace
+	// end = Rparen + 1
+
+	Replace token.Pos
+	Rparen  token.Pos
+
+	Columns []*StarModifierReplaceItem
```
makenowjust, line 657: IMO, we can reuse `Alias` here instead of introducing a new struct.

apstndb: IMO, it is better to separate between `expr AS alias` and `expr [AS] alias`.
Currently they are not compatible with `parseCommaSeparatedList` unless we have a no-arg named `parse*` function.

apstndb: It seems `StarModifierReplace` is the first structure having `expr AS name[, ...]`.
IMO, this is reason enough to have a separate `StarModifierReplaceItem` at this point.

ast/ast.go
```diff
@@ -640,26 +640,71 @@ type SubQuery struct {
...
+}
+
 // Star is a single * in SELECT result columns list.
 //
-//	*
+//	{{"*"}} {{.Except | sqlOpt}} {{.Replace | sqlOpt}}
+//
+// Note: The text/template notation escapes * to avoid normalize * to - by some formatters
+// because the prefix * is ambiguous with bulletin list.
```
makenowjust, line 685: Good catch.

makenowjust (approved):
Thank you!

PR #201, opened 2024-11-13
Refine QueryExpr and add preliminary support of pipe syntax, only FROM, SELECT, WHERE

### LINKED ISSUE #199: Consider about breaking change to support pipe syntax
Pipe syntax is not yet announced in Spanner, but I believe it is on roadmap.
I suspects it requries some breaking change to the current AST nodes.

It may be a good idea to prototype this feature prior to the official release.

### LINKED ISSUE #203: Incompatibility: Paren query expression should able to contain CTE
It is a valid query.

```
$ gcloud spanner databases execute-sql ${SPANNER_DATABASE} --sql '(WITH cte AS (SELECT 1 AS n) SELECT * FROM cte)'
n
1
```

but, memefish can't parse it.

```
$ go run ./tools/parse/main.go --mode query "(WITH cte AS (SELECT 1 AS n) SELECT * FROM cte)"                                                                                
2024/11/13 23:35:51 syntax error: :1:2: expected token: SELECT, but: WITH

  1:  (WITH cte AS (SELECT 1 AS n) SELECT * FROM cte)
       ^~~~
exit status 1
```

## Related syntax

https://cloud.google.com/spanner/docs/reference/standard-sql/query-syntax#sql_syntax

```
query_statement:
  [ statement_hint_expr ]
  [ table_hint_expr ]
  [ join_hint_expr ]
  query_expr

query_expr:
  [ WITH cte[, ...] ]
  { select | ( query_expr ) | set_operation }
  [ ORDER BY expression [{ ASC | DESC }] [, ...] ]
  [ LIMIT count [ OFFSET skip_rows ] ]
```
## Considerations 

* To fix it, it may require breaking change of current fundamental query expressions.
  * It would be related to #199.

### LINKED ISSUE #204: Unsupported SELECT ALL
```sql
SELECT ALL 1
```

It is a valid query.

```
$ gcloud spanner databases execute-sql ${SPANNER_DATABASE} --sql 'SELECT ALL 1'                                                           
(Unspecified)
1
```

But failed in memefish.

```
$ go run github.com/cloudspannerecosystem/memefish/tools/parse@latest --mode query 'SELECT ALL 1'                                         
2024/11/15 07:55:09 syntax error: :1:8: unexpected token: ALL

  1:  SELECT ALL 1
             ^~~
exit status 1
```

## Syntax

```sql
SELECT
  [ { ALL | DISTINCT } ]
  [ AS { typename | STRUCT | VALUE } ]
  select_list
```

Note: I think it is better to introduce node to represent `DISTINCT` or `ALL` or absent(`nil`?) for reusing.
Named `AllOrDistinct`?

apstndb:
Refine QueryExpr to able to support upcoming pipe syntax.

It requires breaking change of top level query structure.

## Breaking changes

- It introduced `ast.Query` and move `WITH`, `ORDER BY`, `LIMIT` clause into it from `ast.QueryStatement` and other `ast.QueryExpr`s.
- ~It introduced `ast.AllOrDistinct` interface, it replaces `distinct` in `ast.Select`.~
- It introduced `ast.AllOrDistinct` typed string. it replaces `distinct` in `ast.Select` and `ast.CompoundQuery`.

## Related issues

- close #199 
- close #203 
- close #204


ast/ast.go
```diff
@@ -97,16 +97,38 @@ func (Insert) isStatement()             {}
...
 	isQueryExpr()
 }
 
 func (Select) isQueryExpr()        {}
+func (Query) isQueryExpr()         {}
+func (FromQuery) isQueryExpr()     {}
 func (SubQuery) isQueryExpr()      {}
 func (CompoundQuery) isQueryExpr() {}
 
+// AllOrDistinct represents ALL or DISTINCT node.
```
makenowjust, line 113: IMO, this should be `type AllOrDistinct string` instead, and place it to `const.go`.

makenowjust: Also, this can be replace a `Distinct` flag in `CompoundQuery`.

apstndb: My thought is that:

* typed string  isn't well compatible with current doc comment and `sql.go` and `ast.go`.
* typed string can't prevent wrong value.

apstndb: Note: I want to reuse it in pipe operators and GQL statements.

makenowjust: It is unclear to me. Please explain it.

> - typed string isn't well compatible with current doc comment, `sql.go`, and `ast.go`.
>   * It spreads logic to many nodes.

No AST node definition is needed to represent one finite type of keyword.
In addition, from my research of the GQL syntax, I am sure that constants of typed strings are sufficient.

Finally, I will make my opinion clear: We MUST use typed string here.

ast/const.go
```diff
@@ -1,5 +1,14 @@
 package ast
 
+type AllOrDistinct string
+
+const (
+	// AllOrDistinctUnspecified is zero value of AllOrDistinct, and it must be ignored in SQL().
```
makenowjust, line 6: Constant names must be `All` and `Distinct`, and zero value is unnecessary.

makenowjust (changes requested):

ast/ast.go
apstndb: I agree it is not required to introduce new AST node type, but it can be a design choice so I haven't yet agree your comment.
We should talk about pros/cons.

> - typed string isn't well compatible with current doc comment, sql.go, and ast.go.
>   - It spreads logic to many nodes.
> It is unclear to me. Please explain it.

1. `SQL()`
Currently, Go `text/format` notation is not needed to be formal, but we need to write `SQL()` equivalent of 
```
{{if .AllOrDistinct}}{{string(.AllOrDistinct)}} {{end}}
```
to many places. It seems `sqlOpt` is easy to write.

2. `Pos()`, `End()`

Actually, `AllOrDistinct` is not necessary to define `pos` and `end` because it does not appear on node boundaries, but let's think about which one we should use if we were to unify either node interface or typed string.
The general pattern for using typed string is that a separate `token.Pos` is required to implement `pos` and `end`, and their code style are not standarized. (Need to store `pos` or `end`?)

https://github.com/cloudspannerecosystem/memefish/blob/bdde76132d427d61b5d32c07cc67e884c66742f5/ast/ast.go#L2146-L2155

https://github.com/cloudspannerecosystem/memefish/blob/bdde76132d427d61b5d32c07cc67e884c66742f5/ast/ast.go#L754-L763

I am doubting that typed string is a good pattern because a single value is no longer self-contained.

apstndb: (It is because `OnDeleteAction` is a constructed by multi token, so we don't use the first token pos, so it seems that we can use `DirEnd` if needed.)

ast/const.go
apstndb: I think `All` is too general a name to bind only `AllOrDistinct`.
If so, how would you name `ALL` in the GQL path search prefix?

https://cloud.google.com/spanner/docs/reference/standard-sql/graph-patterns#search_prefix

```
path_search_prefix:
  {
    ALL
    | ANY
    | ANY SHORTEST
  }
```

apstndb: I'm looking for external guidelines for constants to reference, but I can't find much.
https://google.github.io/styleguide/go/decisions.html#constant-names

> Name constants based on their role, not their values. If a constant does not have a role apart from its value, then it is unnecessary to define it as a constant.
> 
> ```
> // Bad:
> const Twelve = 12
> 
> const (
>     UserNameColumn = "username"
>     GroupColumn    = "group"
> )
> ```

makenowjust: Okay, `AllOrDistinctAll` and `AllOrDistinctDistinct` is accepted.

ast/ast.go
```diff
@@ -101,16 +101,41 @@ func (Insert) isStatement()             {}
...
 	isQueryExpr()
 }
 
 func (Select) isQueryExpr()        {}
+func (Query) isQueryExpr()         {}
+func (FromQuery) isQueryExpr()     {}
 func (SubQuery) isQueryExpr()      {}
 func (CompoundQuery) isQueryExpr() {}
 
+/*
```
makenowjust, line 117: Could you remove this comment out?

makenowjust (changes requested):

ast/ast.go
apstndb: Fixed https://github.com/cloudspannerecosystem/memefish/pull/201/commits/c8a74d98efeee52e1014fa49b14793d4c0fdadcc

makenowjust (approved):
LGTM. Thank you.

PR #207, opened 2024-11-22
Support function hints

### LINKED ISSUE #198: Support function hints
Function hints are not yet supported.

https://cloud.google.com/spanner/docs/reference/standard-sql/functions-reference#function_hints

```
$ go run ./tools/parse --mode query "SELECT
  SUBSTRING(CAST(x AS STRING), 2, 5) AS w,
  SUBSTRING(CAST(x AS STRING), 3, 7) AS y
FROM (SELECT SHA512(z) @{DISABLE_INLINE = TRUE} AS x FROM t)"
2024/11/12 23:35:16 syntax error: :4:24: expected token: ), but: @

  4:  FROM (SELECT SHA512(z) @{DISABLE_INLINE = TRUE} AS x FROM t)
                             ^
exit status 1
```

apstndb:
This PR implements function hints.

Note: The issue is reported by external developer, so it may be better to be prioritized.

## Related issues

- close #198


apstndb:
Need to update `CallExpr.SQL()`.

makenowjust (approved):
Good. Thank you!

Issue #212, opened 2024-11-24
Bug: JoinMethod handling is not correct


apstndb:
## TL;DR

- The current [`JoinMethod`](https://pkg.go.dev/github.com/cloudspannerecosystem/memefish@v0.0.0-20241122062610-75946e1e2219/ast#JoinMethod) is not correct.
- `APPLY` and `LOOP` is not a valid `JOIN` prefixes.

## Description

Only documented `join_method`(not hint) is `HASH`, so `APPLY` and `LOOP` are not valid `join_method`.

```
condition_join_operator:
  {
    [[INNER] [ join_method ] JOIN](https://cloud.google.com/spanner/docs/reference/standard-sql/query-syntax#inner_join)
    | [FULL [OUTER] [ join_method ] JOIN](https://cloud.google.com/spanner/docs/reference/standard-sql/query-syntax#full_join)
    | [LEFT [OUTER] [ join_method ] JOIN](https://cloud.google.com/spanner/docs/reference/standard-sql/query-syntax#left_join)
    | [RIGHT [OUTER] [ join_method ] JOIN](https://cloud.google.com/spanner/docs/reference/standard-sql/query-syntax#right_join)
  }

join_method:
  { HASH }
```

```sql
SELECT * FROM Singers APPLY JOIN Albums USING (SingerId);
SELECT * FROM Singers LOOP JOIN Albums USING (SingerId);
```
are interpreted as
```sql
SELECT * FROM Singers AS APPLY JOIN Albums USING (SingerId);
SELECT * FROM Singers AS LOOP JOIN Albums USING (SingerId);
```

It can be verified using `.*`.

```
spanner> SELECT APPLY.* FROM Singers APPLY JOIN Albums USING (SingerId);
+----------+-----------+----------+------------+------------+
| SingerId | FirstName | LastName | SingerInfo | BirthDate  |
+----------+-----------+----------+------------+------------+
| 1        | Marc      | Richards | NULL       | 1970-09-03 |
| 1        | Marc      | Richards | NULL       | 1970-09-03 |
| 2        | Catalina  | Smith    | NULL       | 1990-08-17 |
| 2        | Catalina  | Smith    | NULL       | 1990-08-17 |
| 2        | Catalina  | Smith    | NULL       | 1990-08-17 |
| 3        | Alice     | Trentor  | NULL       | 1991-10-02 |
| 4        | Lea       | Martin   | NULL       | 1991-11-09 |
[...]

PR #213, opened 2024-11-24
Fix JoinMethod handling

apstndb:
This PR fixes `JoinMethod` handling.
## Breaking changes

This PR only deleted invalid symbols. It can be a treated as a minor breaking change.

## Related issue

- #212


makenowjust (approved):
LGTM. Thank you!

PR #221, opened 2024-12-02
Fix lookaheadLambdaArg not to use parseExpr

apstndb:
This PR fixes lookaheadLambdaArg not to use `parseExpr()`.


makenowjust (approved):
LGTM.

PR #222, opened 2024-12-09
Support vector_length

apstndb:
This PR add support of `ARRAY` with `vector_length`.
The implementation supports any number of named argument compatible structure.
I believe this is a future-proof implementation.

## Breaking changes

- There is no breaking change.


makenowjust (approved):
Good. Thank you.

Issue #190, opened 2024-11-01
Normalize whitespace of SQL()


apstndb:
There are whitespaces that feel unnatural to me.
I think memefish `SQL()` should respect style in official documentation of Spanner except newlines.

TODO: I will add examples.

* Prefer `CREATE INDEX i1 ON t1(col)` than `CREATE INDEX i1 ON t1 (col)`
  * [All examples](https://cloud.google.com/spanner/docs/secondary-indexes?hl=en) use this format.

PR #228, opened 2024-12-18
Fix the format of CREATE INDEX statement

apstndb:
This PR fixes the format of `CREATE INDEX` statement to more official format.

```diff
-CREATE INDEX foo_bar ON foo (bar DESC, baz ASC)
+CREATE INDEX foo_bar ON foo(bar DESC, baz ASC)
```

It is requested by another user in GCPUG Slack `#memefish` so I want to prioritize it.

- Part of #190
  - I think it is not finished because I am not yet research other differences.


makenowjust (approved):
LGTM. Thank you.

PR #230, opened 2024-12-18
Introduce `BadNode` and `MultiError`

makenowjust:
This PR introduces `BadNode` and `MultiError`.

- `BadNode` is a placeholder node for a source fragment containing syntax errors.
- `MultiError` is an error consisting of a list of `*memefish.Error`.
- Now, the parser can report multiple errors and return an AST node even if errors are reported.

Example:

```console
$ go run ./tools/parse/main.go 'select (1 +) * (% 2)'
--- Error
syntax error: :1:12: unexpected token: )
  1|  select (1 +) * (% 2)
   |             ^
syntax error: :1:17: unexpected token: %
  1|  select (1 +) * (% 2)
   |                  ^

--- AST
&ast.QueryStatement{
  Query: &ast.Select{
    Results: []ast.SelectItem{
      &ast.ExprSelectItem{
        Expr: &ast.BinaryExpr{
          Op:   "*",
          Left: &ast.ParenExpr{
            Lparen: 7,
            Rparen: 11,
            Expr:   &ast.BadExpr{
              BadNode: &ast.BadNode{
                NodePos: 8,
                NodeEnd: 11,
                Tokens:  []*token.Token{
                  &token.Token{
                    Kind: "<int>",
                    Raw:  "1",
                    Base: 10,
                    Pos:  8,
                    End:  9,
                  },
                  &token.Token{
                    Kind:  "+",
                    Space: " ",
                    Raw:   "+",
                    Pos:   10,
                    End:   11,
                  },
                },
              },
            },
          },
          Right: &ast.ParenExpr{
            Lparen: 15,
            Rparen: 19,
            Expr:   &ast.BadExpr{
              BadNode: &ast.BadNode{
                NodePos: 16,
                NodeEnd: 19,
                Tokens:  []*token.Token{
                  &token.Token{
                    Kind: "%",
                    Raw:  "%",
                    Pos:  16,
                    End:  17,
                  },
                  &token.Token{
                    Kind:  "<int>",
                    Space: " ",
                    Raw:   "2",
                    Base:  10,
                    Pos:   18,
                    End:   19,
                  },
                },
              },
            },
          },
        },
      },
    },
  },
}

--- SQL
SELECT (1 +) * (% 2)
```


parser.go
```diff
@@ -244,7 +195,18 @@ func (p *Parser) parseStatements(doParse func()) {
 //
 // ================================================================================
 
-func (p *Parser) parseQueryStatement() *ast.QueryStatement {
+func (p *Parser) parseQueryStatement() (stmt *ast.QueryStatement) {
+	l := p.Lexer.Clone()
+	defer func() {
+		if r := recover(); r != nil {
+			// When parsing is failed on tryParseHint or tryParseWith, the result of these methods are discarded
```
apstndb, line 202: I have concerning discarding hints and CTEs means input statement can't be recovered by `SQL()`. I think it don't satisfy users needs(parse-unparse should be safe)

makenowjust: Parsing results are discarded, but the contents (tokens) are preserved. Therefore, your concern is not a problem.

makenowjust: I mean, parse-unparse is still safe after this PR.

apstndb: Thank you for explanation.
I understood all `handleParse*Error` will eat from `l.Token.Pos` to their stop position.

apstndb:
I feel like this specification hasn't been fully discussed, and I have some personal concerns about the current behavior.

As a memefish contributor, it's clear that this change introduces new complexity. I believe there should be a strong enough benefit to justify introducing this complexity.

Furthermore, as a developer of tools that use memefish, I don't find the current specification sufficiently useful. Because Badnode implements too many interfaces, it cannot be used effectively for type assertions. In fact, it has the potential to introduce bugs, so I would likely choose to avoid using BadNode altogether.

My suggestion is to break down the Node into smaller, more specific types for implementing interfaces. If we did this, it would be genuinely more useful than before this PR, and I might actually use it going forward.

PoC branch: https://github.com/cloudspannerecosystem/memefish/compare/feature/bad-node...apstndb:memefish:feature/poc-bad-nodes

Typical code with my usecase
feature/bad-node HEAD https://go.dev/play/p/iOLRu_m1q3F
apstndb/feature/bad-nodes HEAD https://go.dev/play/p/lsXS1D-S8Q3
master HEAD https://go.dev/play/p/OydF15UWANp

apstndb:
Yeah, I may prefer `ParseStatements()` rather than `ParseStatement()` for this usecase.

feature/bad-node HEAD: https://go.dev/play/p/_MuwVgpz50L
apstndb/feature/poc-bad-node HEAD: https://go.dev/play/p/C9aqBbcMGQt

PR #231, opened 2024-12-18
Implement CALL statement

### LINKED ISSUE #152: Support CALL statement
Syntax

https://cloud.google.com/spanner/docs/reference/standard-sql/procedural-language#call

```
CALL procedure_name (procedure_argument[, …])
```

Example

```
$ go run ./tools/parse --mode statement 'CALL cancel_query("12345");'          
2024/10/20 02:01:51 syntax error: :1:1: unexpected token: <ident>

  1:  CALL cancel_query("12345");
      ^~~~
exit status 1

```

See also [Cancel an expensive query](https://cloud.google.com/spanner/docs/introspection/oldest-active-queries#cancel_an_expensive_query).

apstndb:
This PR implements `CALL` statement.

- `CALL` statement is a first neither of query nor DDL nor DML, so it is treated only as `ast.Statement`.
  - `input/statement` directory is added.
- Existing `(*Parser).parseCall()`, which is used for `ast.CallExpr` etc., is renamed to `(*Parser).parseCallLike()` to make more consistent.
- Currently, `CALL` is only used for `cancel_query`, but it is implemented as ZetaSQL compatible for future extension.

## Related issues

- close #152

### DIFF
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -100,6 +100,7 @@ func (DropVectorIndex) isStatement()    {}
 func (Insert) isStatement()             {}
 func (Delete) isStatement()             {}
 func (Update) isStatement()             {}
+func (Call) isStatement()               {}
 
 // QueryExpr represents query expression, which can be body of QueryStatement or subqueries.
 // Select and FromQuery are leaf QueryExpr and others wrap other QueryExpr.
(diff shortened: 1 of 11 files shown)


makenowjust (approved):
LGTM :+1:

Issue #233, opened 2024-12-19
Default pretty printing for specific nodes.


apstndb:
- `CREATE TABLE`
- `CREATE MODEL`

makenowjust:
IMO, "pretty print" is richer/more complex mechanism than this issue expects. Please fix the title to such like "Improve `SQL()` formats for specific AST nodes".

PR #234, opened 2024-12-19
Pretty print only CREATE TABLE

apstndb:
This PR changes `(*CreateTable).SQL()`(and `(*Cluster).SQL()` and `(*CreateRowDeletionPolicy).SQL()`) to pretty print style.

Input( `testdata/input/ddl/create_table_for_format_test.sql`)

```sql

create table if not exists foo (
    foo int64,
    bar float64 not null,
    baz string(255) not null options(allow_commit_timestamp = null),
    qux string(255) not null as (concat(baz, "a")) stored,
    foreign key (foo) references t2 (t2key1),
    constraint fkname foreign key (foo, bar) references t2 (t2key1, t2key2),
    check (foo > 0),
    constraint cname check (bar > 0),
    corge timestamp not null default (current_timestamp())
) primary key (foo),
  interleave in parent foobar,
  row deletion policy ( older_than ( baz, INTERVAL 30 DAY ) )
```

Parse-Unparsed 

```sql
CREATE TABLE IF NOT EXISTS foo (
    foo INT64,
    bar FLOAT64 NOT NULL,
    baz STRING(255) NOT NULL OPTIONS (allow_commit_timestamp = null),
    qux STRING(255) NOT NULL AS (concat(baz, "a")) STORED,
    corge TIMESTAMP NOT NULL DEFAULT (current_timestamp()),
    FOREIGN KEY (foo) REFERENCES t2 (t2key1),
    CONSTRAINT fkname FOREIGN KEY (foo, bar) REFERENCES t2 (t2key1, t2key2),
    CHECK (foo > 0),
    CONSTRAINT cname CHECK (bar > 0)
) PRIMARY KEY (foo),
INTERLEAVE IN PARENT foobar,
ROW DELETION POLICY ( OLDER_THAN ( baz, INTERVAL 30 DAY ))
```

Related issues

- part of #233


makenowjust (approved):
LGTM. Thank you.

PR #237, opened 2024-12-19
Fix literals format

### LINKED ISSUE #236: Emulate string/bytes literal format in FORMAT("%T", input)
## Current behavior

- Currently, string literals and bytes literals are formatted using `QuoteSQLString` and `QuoteSQLBytes`.
- They are always double quoted and double quotes and single quotes are always escaped.

```
$ go run github.com/cloudspannerecosystem/memefish/tools/parse@latest --mode query "SELECT \"'\", \"\\\"\", \"'\\\"\""
...
--- SQL
SELECT "\'", "\"", "\'\""
```

- Bytes literals are formatter as `B""` and upper hex by `"\x%02X`.
```
$ go run github.com/cloudspannerecosystem/memefish/tools/parse@latest --mode expr 'b"あいうえおaiueo"'                
--- SQL
B"\xE3\x81\x82\xE3\x81\x84\xE3\x81\x86\xE3\x81\x88\xE3\x81\x8Aaiueo"
```

## GoogleSQL `FORMAT("%T", input) behavior

In GoogleSQL, one of standard literal format is output of `FORMAT("%T", input)`.

- Literals can be single quoted if there is double quotes and no single quotes, and quotes are not escaped as possible.

```
spanner> SELECT FORMAT("%T", '"') AS has_double, FORMAT("%T", "'") AS has_single, FORMAT("%T", "'\"") AS has_both;
+------------+------------+----------+
| has_double | has_single | has_both |
+------------+------------+----------+
| '"'        | "'"        | "'\""    |
+------------+------------+----------+
```

- Bytes literals use `b""` and lower hex by `\x%02x`.

```
spanner> SELECT FORMAT("%T", CAST("あいうえおaiueo" AS BYTES));
+----------------------------------------------------------------------+
|                                                                      |
+----------------------------------------------------------------------+
| b"\xe3\x81\x82\xe3\x81\x84\xe3\x81\x86\xe3\x81\x88\xe3\x81\x8aaiueo" |
+----------------------------------------------------------------------+
```

- Bytes litrals are also single quoted or double quoted

```
spanner> SELECT FORMAT("%T", b'"') AS has_double, FORMAT("%T", b"'") AS has_single, FORMAT("%T", b"'\"") AS has_both;
+------------+------------+----------+
| has_double | has_single | has_both |
+------------+------------+----------+
| b'"'       | b"'"       | b"'\""   |
+------------+------------+----------+
```

## Typical difference

Typical difference would be JSON literals which have many double quotes.

```
$ go run github.com/cloudspannerecosystem/memefish/tools/parse@latest --mode expr "$(cat <<EOF
JSON '{"foo": "bar"}'
EOF
)"
...
--- SQL
JSON "{\"foo\": \"bar\"}"
```

```
spanner> SELECT FORMAT("%T", JSON '{"foo": "bar"}');
+----------------------+
|                      |
+----------------------+
| JSON '{"foo":"bar"}' |
+----------------------+
```


## FYI

### Print all 7-bit bytes
```
spanner> SELECT FORMAT("%T", CODE_POINTS_TO_BYTES(GENERATE_ARRAY(0, 0x7f))) AS seven_bits_chars;
+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| seven_bits_chars                                                                                                                                                                                                                         |
+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| b"\x00\x01\x02\x03\x04\x05\x06\x07\x08\x09\x0a\x0b\x0c\x0d\x0e\x0f\x10\x11\x12\x13\x14\x15\x16\x17\x18\x19\x1a\x1b\x1c\x1d\x1e\x1f !\"#$%&'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\\]^_`abcdefghijklmnopqrstuvwxyz{|}~\x7f" |
+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
```

### ZetaSQL

- implementation https://github.com/google/zetasql/blob/a516c6b26d183efc4f56293256bba92e243b7a61/zetasql/public/strings.cc#L728-L733
- tests https://github.com/google/zetasql/blob/a516c6b26d183efc4f56293256bba92e243b7a61/zetasql/public/strings_test.cc#L763-L835

apstndb:
This PR align literals format to Spanner/ZetaSQL `FORMAT("%T", input)` functions.

- Generic quoted literals and quoted identifiers
  - `' '` and `'~'` should be printable characters.
  - Use lowercased hex rather than uppercased hex
  - Don't escape quotes except currently used.
  - Only `\` is always escaped printable character.
  - Use single quote if double quote is not suitable.
- String literals
  - Use `\x%02x` for non-printable runes of `r <= 0x7f`.
  - Use only `\t`, `\r`, `\n` in `r < 0x20` instead of `\x%02x`.
- Bytes literals
  - Use `b""` rather than `B""` for bytes literal

- close https://github.com/cloudspannerecosystem/memefish/issues/236


token/quote.go
```diff
@@ -10,30 +10,53 @@ import (
...
 	for _, b := range bs {
-		q := quoteSingleEscape(rune(b))
+		q := quoteSingleEscape(quote, rune(b))
 		if q != "" {
 			buf.WriteString(q)
 			continue
 		}
-		if char.IsPrint(b) {
+
+		// Note: char.IsPrint(' ') is false
```
apstndb, line 52: I think it is better to fix `char.IsPrint`, isn't it?

makenowjust: Yes. You could fix `char.IsPrint`.

makenowjust (approved):
LGTM :+1:

PR #239, opened 2024-12-19
Fix to support ARRAY<PROTO>

### LINKED ISSUE #238: Arrays of protos aren't supported in DDL
Parsing DDL containing a column that's an array of protos fails with an error of the following shape:
```
syntax error: :4:19: expect ident: BOOL, INT64, FLOAT32, FLOAT64, DATE, TIMESTAMP, NUMERIC, JSON, TOKENLIST, STRING, BYTES, but: myproto
	
	  4:  				MyEnums ARRAY<myproto.v1.MyEnum> NOT NULL
	                        ^~~~~~~
```

To replicate the issue:
```
package main

import (
	"github.com/cloudspannerecosystem/memefish"
)

func main() {
	_, err := memefish.ParseDDLs("", `
		CREATE TABLE TestTable (
				ID INT64,
				MyEnums ARRAY<myproto.v1.MyEnum> NOT NULL
		) PRIMARY KEY (ID);
	`)
	if err != nil {
		panic(err)
	}
}
```

The same `CREATE TABLE` statement succeeds without error on Spanner.

Note that using backticks does not resolve the issue.
```
package main

import (
	"github.com/cloudspannerecosystem/memefish"
)

func main() {
	_, err := memefish.ParseDDLs("", `
		CREATE TABLE TestTable (
				ID INT64,
				MyEnums ARRAY<`+"`myproto.v1.MyEnum`"+`> NOT NULL
		) PRIMARY KEY (ID);
	`)
	if err != nil {
		panic(err)
	}
}
```

Error:
```
syntax error: :4:19: expect ident: BOOL, INT64, FLOAT32, FLOAT64, DATE, TIMESTAMP, NUMERIC, JSON, TOKENLIST, STRING, BYTES, but: myproto.v1.MyEnum
	
	  4:  				MyEnums ARRAY<`myproto.v1.MyEnum`> NOT NULL
	                        ^~~~~~~~~~~~~~~~~~~
```

The above testing was performed against commit 10486147cd3c.

apstndb:
This PR fixes the bug that memefish can't parse `ARRAY<PROTO>`.

- `NamedType` is now parsed in `parseScalarSchemaType()`.

## Related issues

- fix #238


makenowjust (approved):
Good catch. Thank you @apstndb and @will-arxed!

PR #248, opened 2024-12-31
Implement inline PRIMARY KEY

### LINKED ISSUE #247: Support inline PRIMARY KEY syntax
https://cloud.google.com/spanner/docs/reference/standard-sql/data-definition-language#create_table
```
CREATE TABLE [ IF NOT EXISTS ] table_name ( [
   { column_name data_type [NOT NULL]
     [ { DEFAULT ( expression ) | AS ( expression ) [ STORED ]
       | GENERATED BY DEFAULT AS IDENTITY [ ( sequence_option_clause ... ) ] } ]
     [ PRIMARY KEY ]
...
 }
   [, ... ]
] ) [ PRIMARY KEY ( [column_name [ { ASC | DESC } ], ...] ) ]
...
```

- `PRIMARY KEY` can be defined as inline column definition.
- `PRIMARY KEY` suffix of `CREATE TABLE` is now optional.

This syntax is already used in many documents so this issue must be prioritized.

https://cloud.google.com/spanner/docs/schema-and-data-model#create_a_parent_table

```sql
CREATE TABLE Singers (
SingerId   INT64 NOT NULL PRIMARY KEY,
FirstName  STRING(1024),
LastName   STRING(1024),
SingerInfo BYTES(MAX),
);
```

apstndb:
This PR implements `PRIMARY KEY` as column definition and make `PRIMARY KEY` suffix of `CREATE TABLE` optional.

There is no breaking changes, but memefish user should know about new `PRIMARY KEY` syntax.

## Related issues

- close #247


makenowjust (approved):
LGTM :+1:

PR #249, opened 2024-12-31
Fix (*FromQuery).SQL() placement

apstndb:
This PR moves misplaced `(*FromQuery).SQL()` into `ast/sql.go` from `ast/ast.go`.


makenowjust (approved):
Good catch. Thank you.

PR #255, opened 2025-01-05
Fix (*ast.Query).End()

apstndb:
This PR fixes `(*ast.Query).End()`. It didn't include after `.Query.End()`.


makenowjust (approved):
LGTM. Thank you!

PR #273, opened 2025-01-18
Remove pp v1 dependency

### LINKED ISSUE #266: Remove k0kubun/pp v1 dependency
We have migrated to `k0kubun/pp/v3`, but there is a remaining dependency.

https://github.com/cloudspannerecosystem/memefish/blob/7a98aad642288b88b1e220b3609c92f753765a27/examples/parse-unparse/main.go#L9

apstndb:
This PR removes k0kubun/pp v1 dependency in examples.

- close #266


makenowjust (approved):
Good catch 👍

PR #275, opened 2025-01-23
Remove unused constants

apstndb:
This PR removes unused constants related to `TableHintKey` and `JoinHintKey`.
Strictly speaking, it is a kind of breaking change because it removes export symbols, but I believe that there is no valid reason that any user code depends on them.


makenowjust (approved):
LGTM 👍 

IMO, this is not a breaking change because there is no reason to use these constants for users.

PR #277, opened 2025-01-28
Support SELECT FOR UPDATE

### LINKED ISSUE #276: Support SELECT FOR UPDATE
https://cloud.google.com/spanner/docs/use-select-for-update

## Syntax

https://cloud.google.com/spanner/docs/reference/standard-sql/query-syntax#sql_syntax

```
query_expr:
  [ WITH cte[, ...] ]
  { select | ( query_expr ) | set_operation }
  [ ORDER BY expression [{ ASC | DESC }] [, ...] ]
  [ LIMIT count [ OFFSET skip_rows ] ]
  [ FOR UPDATE ]
```

## Note

It can be expected to be extended for other lock modes, but `UPDATE` is only valid `LockMode`currently.
https://github.com/google/zetasql/blob/2024.11.1/zetasql/parser/zetasql.tm#L8170-L8182

I think it is better to have `ForUpdate` clause to have pos/end and we can delay to introduce `LockMode` interface later.

apstndb:
This PR implements `FOR UPDATE` clause.

## Related Issue

- close #276

### DIFF
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -668,15 +668,25 @@ type QueryStatement struct {
 // https://cloud.google.com/spanner/docs/query-syntax
 type Query struct {
 	// pos = (With ?? Query).pos
-	// end = (PipeOperators[$] ?? Limit ?? OrderBy ?? Query).end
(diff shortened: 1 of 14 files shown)


makenowjust (approved):
LGTM :+1:

Issue #293, opened 2025-03-22
Support INTERLEAVE IN without PARENT


apstndb:
Spanner now supports `INTERLEAVE IN` without `PARENT`. It doesn't enforce referencial integrity from child to parent.
https://cloud.google.com/spanner/docs/reference/standard-sql/data-definition-language#create_table

```
CREATE TABLE [ IF NOT EXISTS ] table_name ( [
   [, ... ]
] ) [ PRIMARY KEY ( [column_name [ { ASC | DESC } ], ...] ) ]
[, INTERLEAVE IN [PARENT] table_name [ ON DELETE { CASCADE | NO ACTION } ] ]
...
```

It is also supported in `ALTER TABLE` statement.

https://cloud.google.com/spanner/docs/reference/standard-sql/data-definition-language#alter_table

```
ALTER TABLE table_name
    action

where action is:
...
    SET INTERLEAVE IN [ PARENT ] parent_table_name [ ON DELETE { CASCADE | NO ACTION } ]
...
```

apstndb:
Completed via #294 and #295

PR #294, opened 2025-03-22
Support unenforced interleave

apstndb:
This PR implements `INTERLEAVE IN` without `PARENT` in `CREATE TABLE`.
It is implemented as `Enforced bool` flag. ~optional `token.Pos` of `PARENT`.~ 

## Related Issues

- part of #293

### DIFF
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -2660,9 +2660,9 @@ type IndexKey struct {
 	Dir  Direction // optional
 }
 
-// Cluster is INTERLEAVE IN PARENT clause in CREATE TABLE.
+// Cluster is INTERLEAVE IN [PARENT] clause in CREATE TABLE.
 //
-//	, INTERLEAVE IN PARENT {{.TableName | sql}} {{.OnDelete}}
+//	, INTERLEAVE IN {{if .Enforced}}PARENT{{end}} {{.TableName | sql}} {{.OnDelete}}
 type Cluster struct {
 	// pos = Comma
 	// end = OnDeleteEnd || TableName.end
@@ -2671,6 +2671,7 @@ type Cluster struct {
 	OnDeleteEnd token.Pos // end position of ON DELETE clause
 
 	TableName *Path
+	Enforced  bool           // true when PARENT is present
 	OnDelete  OnDeleteAction // optional
 }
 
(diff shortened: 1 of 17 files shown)


apstndb:
I feel `Enforce bool` is better than `token.Pos`.

makenowjust:
The name Enforced sounds confusing in meaning, so I want to ask "what is enforced?"

Is this name derived from ZetaSQL?

apstndb:
https://cloud.google.com/spanner/docs/schema-and-data-model#parent-child
> This parent-child relationship is enforced when interleaved with the `INTERLEAVE IN PARENT` clause.

As written in the documentation, the presence or absence of the `PARENT` keyword indicates whether parent-child referential integrity is **enforced**.

I think this is appropriate if we are to name it based on semantics.

Another way to express it would be like this.

- Presence or absence of the `PARENT` keyword (`bool` or `token.Pos`)
- An `ENUM`-like value corresponding to `IN` or `IN PARENT`
  - This is close to [`INFORMATION_SCHEMA.TABLE.INTERLEAVE_TYPE`](https://cloud.google.com/spanner/docs/information-schema#tables).
  - It is also close to [`ASTSpannerInterleaveClauseEnums.Type`](https://github.com/google/zetasql/blob/a516c6b26d183efc4f56293256bba92e243b7a61/zetasql/parser/ast_enums.proto#L394C9-L400)
    - As explained before, the part published as ZetaSQL and the Spanner DDL parser are completely separate, so it doesn't matter. I don't know when `INTERELEAVE IN` without `PARENT` will be implemented, but `cloud-spanner-emulator` is a better reference..

Since this is an AST, no semantic meaning is needed, so it may be appropriate to define constants `InterleaveTypeIn` and `InterleaveTypeInParent` of type `InterleaveType`.

makenowjust:
If there are variants other than `IN` and `IN PARENT`, I think they should be enum. Otherwise, I have no strong objections, so `Enforced` is fine.

apstndb:
I don't see any reason to add more types than `IN` and `IN PARENT`.
If we use enum instead of bool in case other variants are added in the future, we should stop using bool everywhere else.

makenowjust:
Okay, it is good to go `Enforced` I think.

makenowjust (approved):
LGTM :+1:

PR #313, opened 2025-06-23
Reset errors and allow processing to continue in Parse*

apstndb:
This PR fixes to reset internal errors in `(*memefish.Parser).Parse*()`

Currently, Parse* functions do not reset their internal error state `errors`.
These functions return an error if, after successfully parsing a complete unit/statement (i.e., reaching a final state), the next token encountered is not `token.TokenEOF`.
In other words, if you attempt to parse an `Expr` from the middle of a string (meaning the `Expr` is successfully parsed but is followed by more tokens instead of the expected `token.TokenEOF`), an error is returned. Because this error state is not reset, any subsequent attempts to parse further from the remaining string will also result in an error.
This means these functions cannot be reused to parse multiple fragments from a single string unless their accumulated errors are explicitly ignored.

Current usecase:
https://github.com/apstndb/spanner-mycli/blob/df6f18c1803d8a6570ad78c3bc7cbcd2c2df6371/statements_split_points.go#L243-L263

### DIFF
--- a/parser.go
+++ b/parser.go
@@ -24,7 +24,11 @@ func (p *Parser) ParseStatement() (ast.Statement, error) {
 	}
 
 	if len(p.errors) > 0 {
-		return stmt, MultiError(p.errors)
+		// Reset the errors and allow processing to continue
+		err := MultiError(p.errors)
+		p.errors = nil
+
+		return stmt, err
 	}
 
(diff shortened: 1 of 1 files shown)


makenowjust (approved):
Thank you :+1:

PR #322, opened 2025-09-06
fix: Preserve all comments in SplitRawStatements

apstndb:
```
// SplitRawStatements splits an input string to statement strings at terminating semicolons without parsing.
// It preserves all comments.
```
 
In the current implemention, comments after semicolons are wrongly stripped.
This PR fixes this bug.


makenowjust (approved):
Good 👍 Thank you.

PR #338, opened 2026-02-09
feat: support anonymous table constraint style PRIMARY KEY

apstndb:
Support anonymous `PRIMARY KEY` style as a table constraint inside the parentheses of `CREATE TABLE`.

This is the third way to define a primary key in Spanner, in addition to the inline column definition (`PK INT64 PRIMARY KEY`) and the trailing clause (`PRIMARY KEY (PK)`) outside the table elements. This style is supported by the Spanner emulator and actual database behavior.

## Design decision

Introduced `ast.TablePrimaryKey` as a `TableConstraint` and updated `Parser.parseCreateTable` to support anonymous `PRIMARY KEY` inside the table elements while intentionally excluding named constraints.


makenowjust:
Where is this feature documented? Or, is it an undocumented feature?

apstndb:
As a matter of fact, this is an undocumented feature in Spanner.

It is mentioned in the release notes for cloud-spanner-emulator v1.5.48, and there are also test cases for it.

https://github.com/GoogleCloudPlatform/cloud-spanner-emulator/releases/tag/v1.5.48
> Support specifying PRIMARY KEY as a table constraint

In addition, real Spanner instances accept this syntax. However, it is still not documented in the official Spanner documentation.

makenowjust:
Great. Thank you.

makenowjust (approved):

Issue #298, opened 2025-04-05
Proposal: Unify DDL object names from Ident to Path


apstndb:
Currently, only a limited number of Spanner schema objects support named schemas.

> - `TABLE`
>   -  `CREATE`
>   - `INTERLEAVE IN [PARENT]`
>   - `FOREIGN KEY`
>   - `SYNONYM`
> - `VIEW`
> - `INDEX`
> - `FOREIGN KEY`
> - `SEQUENCE`

However, the parser already handles even unsupported objects as `qualified_identifier`. This suggests it's simply a matter of implementation rather than a parsing limitation.

https://github.com/GoogleCloudPlatform/cloud-spanner-emulator/blob/master/backend/schema/parser/ddl_parser.jjt

Considering consistency, it would be preferable to allow parsing of any valid syntax, as this would minimize the changes required when these features are implemented in the future.

1. For future DDL implementations, represent names using `ast.Path`.
2. For existing DDL as well, replace `ast.Ident` with `ast.Path`.

Issue #98, opened 2024-09-13
Support Spanner Graph


apstndb:
[Spanner Graph](https://cloud.google.com/blog/products/databases/announcing-spanner-graph?hl=en) introduces ISO GQL and SQL/PGQ subsets to Cloud Spanner.
It is coexists with queries and DDLs of GoogleSQL, so I think it is better to be implemented by memefish.

* [`{CREATE | DROP} PROPERTY GRAPH`](https://cloud.google.com/spanner/docs/reference/standard-sql/graph-schema-statements?hl=en#gql_create_graph) DDL statements
* [`GRAPH_TABLE` operator](https://cloud.google.com/spanner/docs/reference/standard-sql/graph-sql-queries?hl=en#graph_table_operator)
* [`IN { gql }`](https://cloud.google.com/spanner/docs/graph/insert-update-delete-data?hl=en#update-nodes-or-edges) syntax
  ```
     -- Use Graph pattern matching to identify Account nodes to update:
  UPDATE Account SET is_blocked = false
  WHERE id IN {
    GRAPH FinGraph
    MATCH (a:Account WHERE a.id = 1)-[:TRANSFERS]->{1,2}(b:Account)
    RETURN b.id
  }
  ```
* [Top level GQL query syntax](https://cloud.google.com/spanner/docs/reference/standard-sql/graph-query-statements?hl=en#gql_syntax)

PR #351, opened 2026-04-22
Support schema-qualified names in PROPERTY GRAPH elements

tykid16:
This PR adds support for schema-qualified (dotted) table names in `CREATE PROPERTY GRAPH` NODE/EDGE TABLES and REFERENCES clauses.

Before this change, schema-qualified names like `myschema.MyTable` in a NODE TABLES clause caused a parse error: `Expected ) or , but got .`.

**Breaking changes**: The types of the following fields are changed to `*ast.Path` from `*ast.Ident`:
- `PropertyGraphElement.Name`
- `PropertyGraphSourceKey.ElementReference`
- `PropertyGraphDestinationKey.ElementReference`

## Local verification

- `go test ./...` — all tests pass
- `make lint` — 0 issues
- `make gen` — generated files up to date

## Related

- #98
- #298

### DIFF
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -3932,7 +3932,7 @@ type PropertyGraphElement struct {
 	// pos = Name.pos
 	// end = (DynamicProperties ?? DynamicLabel ?? Properties ?? Keys ?? Alias ?? Name).end
 
-	Name              *Ident
+	Name              *Path
 	Alias             *Ident                          // optional
 	Keys              PropertyGraphElementKeys        // optional
 	Properties        PropertyGraphLabelsOrProperties // optional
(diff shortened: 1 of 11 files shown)


tykid16:
Closing this PR — further investigation revealed that the root cause is not in memefish but in the Spanner API itself.

Spanner's `CREATE PROPERTY GRAPH` does not support referencing tables in a named schema:
- Schema-qualified names (`irbank.Disclosure`) → parse error from the Spanner API
- Unqualified names (`Disclosure`) → "Table not found" when tables are in a named schema

This is a Spanner-side limitation, not a memefish parsing issue. The memefish parser change proposed here would not resolve the underlying problem.

Apologies for the noise.

PR #375, opened 2026-07-09
fix: drop duplicate "syntax error:" prefix in parseFromQuery

apstndb:
## Summary
- `parseFromQuery` baked a literal `syntax error: ` into ORDER BY / LIMIT rejection messages, but `Error.Error()` already prepends that prefix, producing `syntax error: syntax error: ...`.
- Remove the literal prefix so these messages match every other `errorfAtToken` call site.
- Add `!bad_` goldens for `FROM t ORDER BY x` and `FROM t LIMIT 1`.

## Test plan
- [x] `go run ./tools/parse 'FROM t ORDER BY x'` shows a single `syntax error:` prefix


apstndb:

PR #376, opened 2026-07-09
fix: avoid panic in token.QuoteSQLIdent("")

apstndb:
## Summary
- `needQuoteSQLIdent` read `s[0]` without a length check, so `QuoteSQLIdent("")` panicked with an index-out-of-range.
- Any programmatically built AST with an empty `Ident` therefore crashed on `SQL()`.
- Guard the empty string: an empty identifier cannot be written unquoted, so it is backquoted (``).

## Test plan
- [x] Unit test covers `QuoteSQLIdent("")` → "``"

### DIFF
--- a/token/quote.go
+++ b/token/quote.go
@@ -112,6 +112,11 @@ func quoteSingleEscape(r, quote rune, isString bool) string {
 }
 
 func needQuoteSQLIdent(s string) bool {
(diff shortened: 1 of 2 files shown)


apstndb:

apstndb:
Closing again after verifying the head-branch restoration and update sequence; no review requested in this temporary reopen.

PR #377, opened 2026-07-09
fix: use comma separator in StarModifierExcept.SQL()

apstndb:
## Summary
- `StarModifierExcept.SQL()` joined columns with a space instead of `", "`, so a multi-column EXCEPT such as `SELECT * EXCEPT (a, b)` unparsed to the invalid `EXCEPT (a b)`.
- The sibling `StarModifierReplace` already uses `", "`.
- Add a multi-column golden; every existing EXCEPT golden used exactly one column, so the round-trip check never exercised the separator.

## Test plan
- [x] New golden `select_star_except_multi.sql` round-trips to `EXCEPT (fruit, vegetable)`

### DIFF
--- a/ast/sql.go
+++ b/ast/sql.go
@@ -250,7 +250,7 @@ func (s *SubQuery) SQL() string {
 	return "(" + s.Query.SQL() + ")"
 }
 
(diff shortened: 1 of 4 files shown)


apstndb:

PR #378, opened 2026-07-09
fix: recover first-token lexical errors in Parse* entry points

apstndb:
## Summary
- All public `Parse*` methods called `nextToken()` before any recover was established, so a lexical error in the very first token (unclosed string, illegal byte, etc.) panicked out of the `(node, error)` API.
- The same error after the first token was already recovered — the public contract was violated only for first-token cases (common in REPL/editor input such as a lone `'`).
- Move the initial `nextToken` inside a recover boundary for every entry point, return `Bad*` nodes, and advance once in `noPanic` mode when the restored lexer still holds the zero token so recovery skip loops do not emit a spurious empty token.
- Add `!bad_` goldens covering expr/query/ddl/dml.

## Test plan
- [x] `go run ./tools/parse -mode expr "'hello"` returns error (no panic)
- [x] `go run ./tools/parse -mode expr "1 + 'x"` still returns error (regression)
- [x] Normal parses (`1 + 2`, etc.) unchanged


apstndb:

Issue #268, opened 2025-01-12
Split GQL related PRs


apstndb:
Spanner Graph PR #99 is too large to review, so I want to split this PR into more fine grained PRs.

For example

- Basic structure of Graph Query (`GRAPH`, `NEXT`, `RETURN`, and set operations)
- Other GQL statements
- GQL Subqueries
- Expressions

PR #379, opened 2026-07-09
feat: support GQL MATCH and graph patterns

apstndb:
This PR adds support for the GQL `MATCH` statement:

- `[OPTIONAL] MATCH` statements
- Full graph patterns
- First-class standalone graph-pattern parsing through both
  `memefish.ParseGQLGraphPattern` and `(*Parser).ParseGQLGraphPattern`
- Typed recovery through `ast.GQLGraphPatternNode`: valid input returns
  `*ast.GQLGraphPattern`, while malformed input returns a safe
  `*ast.BadGQLGraphPattern` together with the parse error
- `gql` and `gql_graph_pattern` support in `tools/parse`

The recovered graph-pattern node is safe to unparse, inspect positions, and
walk. It preserves malformed tokens materialized by the lexer, including
first-token string-literal errors. Recovery for a whole GQL graph query remains
separate and is tracked by #383.

## Validation

- `make ci`
- Checked against the Cloud Spanner graph-pattern reference, current
  `google/googlesql` grammar and parser tests, and the embedded Spanner
  emulator:
  - path variables compose with path search prefixes
  - prefix-like non-reserved keywords can be path variable names
  - edge-only subpaths such as `(->)` are accepted
  - empty property filters such as `(n {})` are syntax errors

## Related Issues

- Part of #268
- Related to #383


ast/ast.go
```diff
@@ -4812,3 +4816,339 @@ type GQLOrderByItem struct {
...
+func (BadGQLGraphPattern) isGQLGraphPatternNode() {}
+
+// GQLMatch represents a MATCH statement in GQL.
+//
+//	{{if .OptionalPos.Invalid | not}}OPTIONAL {{end}}MATCH{{if .Hint}} {{.Hint | sql}}{{end}} {{.Pattern | sql}}
+type GQLMatch struct {
+	// pos = OptionalPos || Match
+	// end = Pattern.end
+
+	OptionalPos token.Pos // position of "OPTIONAL", optional
```
makenowjust: `Optional` seems better. But, I have no strong opinion.

```suggestion
	Optional token.Pos // position of "OPTIONAL", optional
```

apstndb: Addressed in d3eb45a: renamed GQLMatch.OptionalPos to Optional and regenerated the generated position code and goldens, matching the existing token.Pos field-naming convention.

makenowjust (approved):
Thanks for implementing a big feature!

makenowjust (approved):

PR #381, opened 2026-07-09
fix: use comma separator in StarModifierExcept.SQL()

### LINKED ISSUE #380: `StarModifierExcept.SQL()` unparses multi-column EXCEPT into invalid SQL
#### Summary

When unparsing an AST with a multi-column EXCEPT clause such as `SELECT * EXCEPT (a, b) FROM t`, `SQL()` produces **invalid SQL** with space-separated columns instead of comma-separated ones.

#### Reproduction steps (verified)

```
$ go run ./tools/parse --output unparse 'SELECT * EXCEPT (a, b) FROM t'
--- SQL
SELECT * EXCEPT (a b) FROM t
```

#### Expected behavior

```
SELECT * EXCEPT (a, b) FROM t
```

#### Notes

Existing golden tests only cover EXCEPT with a single column, so round-trip verification did not catch this. A test case with multiple columns is needed.

apstndb:
This PR fixes `SQL()` of `StarModifierExcept` (`SELECT * EXCEPT (a, b, ...)`)

- fixes #380

### DIFF
--- a/ast/sql.go
+++ b/ast/sql.go
@@ -250,7 +250,7 @@ func (s *SubQuery) SQL() string {
 	return "(" + s.Query.SQL() + ")"
 }
 
-func (s *StarModifierExcept) SQL() string { return "EXCEPT (" + sqlJoin(s.Columns, " ") + ")" }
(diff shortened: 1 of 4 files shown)


makenowjust (approved):
Good catch! Thank you!

Issue #383, opened 2026-07-12
Missing BadGQLGraphQuery


apstndb:
Other public parser families can return a non-nil recovered AST containing a `Bad*` node. `(*Parser).ParseGQLQuery` instead returns `(nil, err)` because there is no `BadGQLGraphQuery`.
I believe It is better to add `BadGQLGraphQuery`.

PR #384, opened 2026-07-12
fix: recover lexer errors at Parse entry boundaries

apstndb:
## Summary

- supersede the closed prototype in #378 with the same focused recovery fix on current `main`;
- establish recovery before the initial token advance in every public `Parser.Parse*` entry point;
- recover lexical errors raised while advancing after semicolons in `ParseStatements`, `ParseDDLs`, and `ParseDMLs`;
- return existing safe `Bad*`/partial AST nodes where supported and drain accumulated errors so a `Parser` can be reused;
- cover all 11 public parser methods, all 10 package-level helpers, post-semicolon recovery, parser reuse, and `SQL`/`Pos`/`End`/`Walk` safety.

## Motivation

memefish documents error recovery as returning an error together with a partial AST containing `Bad*` nodes where possible. However, public parser entry points advanced the lexer before installing their recovery boundary. A lexical error in the first token therefore escaped as a panic instead of returning through the `(AST, error)` API.

The multi-statement entry points had the same boundary gap when advancing after a semicolon. This change keeps every token advance owned by those public entry points inside recovery.

`ParseGQLQuery` continues to return `(nil, err)` because there is currently no `BadGQLGraphQuery`; that additive AST design is tracked separately in #383.

## Compatibility

This changes malformed-input behavior from a panic to an ordinary returned error. It does not add or change public API types, and valid parsing behavior is unchanged.

## Validation

- `make ci`


makenowjust (approved):
Good fix. Thank you!

PR #390, opened 2026-07-20
fix: avoid comment prefix for consecutive unary minus

### LINKED ISSUE #382: Consecutive unary minus becomes a comment prefix after unparse
`- -1` is a valid expression.
```
spanner> SELECT - -1;
+---+
|   |
+---+
| 1 |
+---+
```

But, memefish unparses it as `--1`. It will be parsed as a comment prefix and not roundtrip safe.

```
$ go run ./tools/parse -- 'SELECT - -1'
--- AST
&ast.QueryStatement{
  Query: &ast.Select{
    Results: []ast.SelectItem{
      &ast.ExprSelectItem{
        Expr: &ast.UnaryExpr{
          OpPos: 7,
          Op:    "-",
          Expr:  &ast.IntLiteral{
            ValuePos: 9,
            ValueEnd: 11,
            Base:     10,
            Value:    "-1",
          },
        },
      },
    },
  },
}
--- SQL
SELECT --1
```

## Related GoogleSQL implementation

https://github.com/google/googlesql/blob/36dd14aa0657ea299725504bc0f938732f58f380/googlesql/parser/unparser.h#L92-L96

https://github.com/google/googlesql/blob/36dd14aa0657ea299725504bc0f938732f58f380/googlesql/parser/unparser.cc#L2733-L2749

apstndb:
## Summary

- prevent `UnaryExpr.SQL()` from combining consecutive minus tokens into a `--`
  comment prefix;
- insert a space only when a unary-minus operand renders with a leading minus;
- preserve existing output for other operators;
- cover the parser-produced signed-integer case that reproduced the issue.

## Examples

- `- -1` remains `- -1`, rather than becoming `--1`.

## Relates Issue

- Fixes #382


makenowjust (approved):
Good catch. Thank you.

PR #393, opened 2026-07-22
feat: support GRAPH_TABLE operator

apstndb:
## Summary

- support `GRAPH_TABLE` as a table expression containing either an advanced GQL
  query block or the SQL/PGQ single-`MATCH` form;
- support the single-`MATCH` form both with an optional `COLUMNS` projection and
  without `COLUMNS`;
- support qualified graph names, joins, aliases, and `TABLESAMPLE` suffixes;
- add parser, unparser, position, walking, and round-trip goldens.

## Note

Spanner's GoogleSQL dialect accepts an SQL/PGQ-style single-`MATCH` form
with optional `COLUMNS`, although this form is not currently documented in the
Spanner documentation. The upstream GoogleSQL compliance tests also cover this
grammar:

https://github.com/google/googlesql/blob/1f8aa333f4d6353cd3a64471fc83121df72df3f7/googlesql/compliance/testdata/graph_table.test

```
spanner> SELECT a_id FROM GRAPH_TABLE(FinGraph MATCH (a) COLUMNS(a.id AS a_id));
+------+
| a_id |
+------+
| 7    |
| 16   |
| 20   |
| 1    |
| 2    |
| 3    |
+------+
6 rows in set (9.03 msecs)

spanner> SELECT a.id FROM GRAPH_TABLE(FinGraph MATCH (a));
+----+
| id |
+----+
| 7  |
| 16 |
| 20 |
| 1  |
| 2  |
| 3  |
+----+
6 rows in set (8.82 msecs)
```

## Related Issues

- #98 
- #268


ast/ast.go
```diff
@@ -4659,6 +4660,55 @@ func (GQLLimit) isGQLPrimitiveQueryStatement()                      {}
...
+	Node
+	isGraphTableQuery()
+}
+
+func (GQLMultiLinearQueryStatement) isGraphTableQuery() {}
+func (GraphTableMatch) isGraphTableQuery()              {}
+
+// GraphTableExpr represents GRAPH_TABLE operator in SQL.
+//
+//	GRAPH_TABLE({{.GraphName | sql}} {{.Query | sql}}){{sqlOpt " " .As ""}}{{sqlOpt " " .Sample ""}}
```
makenowjust: I'm not sure this `sqlOpt` usage is valid.

apstndb: ~Fixed in 19046c4. The syntax template now uses standard {{if ...}} blocks for the optional alias and TABLESAMPLE suffixes, matching the pseudo-template vocabulary documented at the top of ast/ast.go.~

apstndb: Follow-up in 6b80515: aligned this with the established doc-template convention by using the one-argument pipeline form (`{{.As | sqlOpt}}` and `{{.Sample | sqlOpt}}`). The three-argument form belongs only to the Go helper in ast/sql.go.

makenowjust: Ah, I see. We might want to fill the gap between the doc template and its implementaion.

PR #394, opened 2026-07-22
feat: support GQL graph operators

apstndb:
This PR adds support for [GQL graph operators](https://docs.cloud.google.com/spanner/docs/reference/standard-sql/graph-operators).

## Summary

- support `IS [NOT] SOURCE [OF]`, `IS [NOT] DESTINATION [OF]`, and
  `IS [NOT] LABELED` expressions;
- preserve graph-label expression precedence when parsing and unparsing;
- add expression and GQL `WHERE` golden coverage, including the documented
  graph concatenation and function-style graph predicates.

The function-style graph predicates `ALL_DIFFERENT`, `PROPERTY_EXISTS`, and
`SAME` already parse through the existing function-call AST and need no
dedicated expression nodes.

## Related Issues

- #98 
- #268


makenowjust (approved):
Looks good :+1:

PR #399, opened 2026-08-11
feat: support IS UNKNOWN expressions

apstndb:
This PR adds support for `IS [NOT] UNKNOWN` expression.


ast/ast.go
```diff
@@ -1438,6 +1439,19 @@ type IsBoolExpr struct {
 	Right bool
 }
 
+// IsUnknownExpr is IS UNKNOWN expression node.
+//
+//	{{.Left | sql}} IS {{if .Not}}NOT{{end}} UNKNOWN
```
makenowjust, line 1444: Is this better?

```suggestion
//	{{.Left | sql}} IS{{if .Not}} NOT{{end}} UNKNOWN
```

apstndb: I think the current form is preferable as documentation. These templates are not intended to specify whitespace exactly, and keeping spaces outside the template actions makes the SQL token boundaries easier to read. Tight adjacency is useful when tokens are intentionally concatenated, such as `SAFE_` and `CAST`, but that is not the case here. This also matches the nearby `IsNullExpr` and `IsBoolExpr` templates.

makenowjust (approved):
Thanks @apstndb!

PR #407, opened 2026-08-16
fix: avoid panic in token.QuoteSQLIdent("")

### LINKED ISSUE #406: Avoid a panic in QuoteSQLIdent for empty input
## Problem

`token.QuoteSQLIdent("")` panics with an index-out-of-range error because `needQuoteSQLIdent` reads `s[0]` without first checking whether `s` is empty.

```text
panic: runtime error: index out of range [0] with length 0
```

The same panic is reachable indirectly through an exported AST node:

```go
_ = (&ast.Ident{}).SQL()
```

### `token.QuoteSQLIdent("")`
https://go.dev/play/p/vniPESiuWGD

```go
package main

import "github.com/cloudspannerecosystem/memefish/token"

func main() {
	_ = token.QuoteSQLIdent("")
}
```

### `(*ast.Ident).SQL()`

https://go.dev/play/p/Jp6w9qpqLfE

```go
package main

import "github.com/cloudspannerecosystem/memefish/ast"

func main() {
	_ = (&ast.Ident{Name: ""}).SQL()
}
```

## Why change this

The function returns only a string and documents no non-empty precondition. Downstream code that constructs `ast.Ident` values from runtime strings must currently validate every identifier before serialization or recover at a broader boundary.

This follows the separation in `google/googlesql`: `ParseIdentifier` rejects an empty backquoted identifier, while `ToIdentifierLiteral("")` returns two backticks and [explicitly notes](https://github.com/google/googlesql/blob/3eec7971a4706b04c10019f89d09dda854442c30/googlesql/public/strings_test.cc#L796-L797) that the result is not a valid identifier.

## Expected behavior

`QuoteSQLIdent("")` returns two backticks instead of panicking. Identifier validation remains unchanged; empty identifiers are still rejected by the lexer/parser.

apstndb:
## Summary
- Make `QuoteSQLIdent("")` return two backticks instead of panicking.
- Add regression coverage for the empty input.

## Related Issues

- fixes #406

### DIFF
--- a/token/quote.go
+++ b/token/quote.go
@@ -112,6 +112,11 @@ func quoteSingleEscape(r, quote rune, isString bool) string {
 }
 
 func needQuoteSQLIdent(s string) bool {
+	// An empty identifier cannot be written without backquotes.
+	if s == "" {
+		return true
+	}
+
 	// When s is keyword, it should be quoted.
 	if IsKeyword(s) {
 		return true
--- a/token/quote.go
+++ b/token/quote.go
@@ -112,7 +112,7 @@ func quoteSingleEscape(r, quote rune, isString bool) string {
 }
 
 func needQuoteSQLIdent(s string) bool {
-	// An empty identifier cannot be written without backquotes.
+	// QuoteSQLIdent formats identifiers without validating them.
 	if s == "" {
 		return true
 	}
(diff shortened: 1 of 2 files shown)


makenowjust (approved):
:+1:

PR #422, opened 2026-09-03
Reject braced constructor field without value

### LINKED ISSUE #418: bug: `ParseExpr` accepts `NEW A00{A0}` without error, but SQL() panics with nil pointer dereference
Found by fuzzing (see #415).

## Repro

```go
expr, err := memefish.ParseExpr("", "NEW A00{A0}")
// err == nil
_ = expr.SQL() // panics
```

```
panic: runtime error: invalid memory address or nil pointer dereference

github.com/cloudspannerecosystem/memefish/ast.(*BracedConstructorField).SQL(...)
        ast/sql.go:736
github.com/cloudspannerecosystem/memefish/ast.(*BracedConstructor).SQL(...)
        ast/sql.go:727
github.com/cloudspannerecosystem/memefish/ast.(*BracedNewConstructor).SQL(...)
        ast/sql.go:723
```

## Cause

The parser accepts a braced NEW constructor field without a value (`{A0}`) and returns `err == nil`, producing a `BracedConstructorField` whose value is nil; `BracedConstructorField.SQL` then dereferences it.

Either the parser should reject a field with no value (per the [braced constructor grammar](https://cloud.google.com/spanner/docs/reference/standard-sql/operators#braced_constructors), a field requires a value), or the AST invariant should allow the nil and `SQL()` should handle it. Rejecting at parse time seems right.

makenowjust:
Fixes #418.

`parseBracedNewConstructorField` switched on `:` and `{` but had no `default` case, so a field followed by anything else (e.g. `NEW A00{A0}`, where the field is followed by `}`) produced a `BracedConstructorField` with a nil `Value` — and `err == nil`. `SQL()` and the position accessors then dereferenced the nil value and panicked.

A braced constructor field requires a value, so the parser now reports it as a syntax error:

```
syntax error: :1:15: expected token: {, :, but: }
  1|  NEW foo { bar }
   |                ^
```

Error recovery wraps the expression in an `ast.BadExpr` as usual.

## Tests

- New testdata case `testdata/input/expr/!bad_new_braced_constructor_field_without_value.sql` (`NEW foo { bar }`) with its generated result, alongside the existing `!bad_new_braced_constructor.sql`.

### DIFF
--- a/testdata/result/expr/!bad_new_braced_constructor_field_without_value.sql.txt
+++ b/testdata/result/expr/!bad_new_braced_constructor_field_without_value.sql.txt
@@ -1,0 +1,51 @@
+--- !bad_new_braced_constructor_field_without_value.sql
+NEW foo { bar }
+
+--- Error
+syntax error: testdata/input/expr/!bad_new_braced_constructor_field_without_value.sql:1:15: expected token: {, :, but: }
+  1|  NEW foo { bar }
+   |                ^
+syntax error: testdata/input/expr/!bad_new_braced_constructor_field_without_value.sql:1:15: expected token: <eof>, but: }
+  1|  NEW foo { bar }
+   |                ^
+
+
+--- AST
+&ast.BadExpr{
+  BadNode: &ast.BadNode{
+    NodeEnd: 13,
+    Tokens:  []*token.Token{
+      &token.Token{
+        Kind: "NEW",
+        Raw:  "NEW",
+        End:  3,
+      },
+      &token.Token{
+        Kind:     "<ident>",
+        Space:    " ",
+        Raw:      "foo",
+        AsString: "foo",
+        Pos:      4,
+        End:      7,
+      },
+      &token.Token{
+        Kind:  "{",
+        Space: " ",
+        Raw:   "{",
+        Pos:   8,
+        End:   9,
+      },
+      &token.Token{
+        Kind:     "<ident>",
+        Space:    " ",
+        Raw:      "bar",
+        AsString: "bar",
+        Pos:      10,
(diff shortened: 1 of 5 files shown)


makenowjust:
@apstndb Could you review this?

apstndb (approved):
LGTM

testdata/result/expr/!bad_new_braced_constructor_field_without_value.sql.txt
```diff
@@ -0,0 +1,51 @@
+--- !bad_new_braced_constructor_field_without_value.sql
+NEW foo { bar }
+
+--- Error
+syntax error: testdata/input/expr/!bad_new_braced_constructor_field_without_value.sql:1:15: expected token: {, :, but: }
+  1|  NEW foo { bar }
+   |                ^
+syntax error: testdata/input/expr/!bad_new_braced_constructor_field_without_value.sql:1:15: expected token: <eof>, but: }
+  1|  NEW foo { bar }
```
makenowjust, line 9: This error looks strange. We want to avoid it.

testdata/result/expr/!bad_new_braced_constructor_field_without_value.sql.txt
```diff
@@ -0,0 +1,51 @@
...
+        AsString: "bar",
+        Pos:      10,
+        End:      13,
+      },
+    },
+  },
+}
+
+--- SQL
+NEW foo { bar
```
makenowjust, line 51: Furthermore, this lacks the closing brace `}`. It seems actually a bug.

PR #429, opened 2026-09-06
fix: correct CREATE MODEL IF NOT EXISTS order

### LINKED ISSUE #428: bug: CREATE MODEL IF NOT EXISTS uses the wrong clause order
The [Spanner DDL reference](https://docs.cloud.google.com/spanner/docs/reference/standard-sql/data-definition-language#create_model) places `IF NOT EXISTS` before the model name:

```sql
CREATE MODEL IF NOT EXISTS MyModel REMOTE
```

Currently, `ParseDDL` rejects this at `IF`. Instead, it accepts and emits the invalid order:

```
$ go run github.com/cloudspannerecosystem/memefish/tools/parse@latest -output unparse 'CREATE MODEL IF NOT EXISTS MyModel REMOTE'
--- Error
syntax error: :1:14: expected token: <ident>, but: IF
  1|  CREATE MODEL IF NOT EXISTS MyModel REMOTE
   |               ^~

--- SQL
CREATE MODEL IF NOT EXISTS MyModel REMOTE
```

```
$ go run github.com/cloudspannerecosystem/memefish/tools/parse@latest -output unparse 'CREATE MODEL MyModel IF NOT EXISTS REMOTE'
--- SQL
CREATE MODEL MyModel IF NOT EXISTS REMOTE
```

Managed Spanner verification on September 6, 2026 confirmed that the correct order executes with a configured endpoint, while the name-first order is a syntax error.

The parser and `SQL()` should use the documented order. `CREATE OR REPLACE MODEL` remains a separate alternative and must not accept `IF NOT EXISTS` as well.

apstndb:
- Fixes #428 

Parse and emit `IF NOT EXISTS` before the model name. Reject the misplaced clause and its combination with `OR REPLACE`.

Add goldens for valid and invalid forms.


makenowjust (approved):
:+1:

